PR #7, opened 2019-11-18
RTP Demo

mat-hek:


kanes115:
Can you push the history from the other repo?

mat-hek:
@kanes115 no idea how

rtp/mix.exs
```diff
@@ -0,0 +1,36 @@
...
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      # {:dep_from_hexpm, "~> 0.3.0"},
+      # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
+      {:membrane_core, [env: :prod, path: "~/membrane-core", override: true]},
+      {:membrane_bin_rtp, path: "~/membrane_bin_rtp"},
```
derekclair: It seems to be exactly the right direction, and while I'd happily provide more in-depth feedback; _this_ dependency isn't published so I can't run this.

kanes115: It is going to be published along with Membrane v0.5 which is happening this week.

PR #8, opened 2019-11-19
Add bin demo as submodule

kanes115:


kanes115:
Closing due to #7 and the fact we don't want to have a separate repo for rtp demo.

PR #24, opened 2020-11-27
Change element_file -> file_plugin. Update other deps 

mickel8:


basic/mix.exs
```diff
@@ -21,19 +21,18 @@ defmodule Membrane.Demo.Basic.MixProject do
   # Run "mix help compile.app" to learn about applications.
   def application do
     [
-      extra_applications: [:logger]
+      extra_applications: []
```
mat-hek, line 24: why? It seems logger is not added by default

mickel8: Yep, that's true, I missed this with config in which there was old logger configuration and thought we have to remove this too

PR #26, opened 2020-12-22
Basic

mat-hek:


mickel8 (commented):
I am wondering if we should split these examples into two separate elixir projects and name them `basic_pipeline` and `basic_element` or something like this. The point is that at the first glance it can be hard for newbie to see the difference between these two examples (we run them almost in the same way and they both play mp3). Also we don't write which source code corresponds to which example. This is obvious for us but I am not sure if for someone new too

mickel8 (approved):

PR #28, opened 2020-12-30
update readme

mat-hek:


mickel8 (commented):
What do you think about providing similar README for webrtc subdirectory?

mat-hek:
@mickel8 maybe let's decide when we finish other webrtc examples

mickel8 (approved):

PR #29, opened 2021-01-05
Fix hardcoded SDP offer

mickel8:


mickel8:
echo demo has been replaced by videoroom demo in #30

PR #30, opened 2021-01-20
Videoroom

mat-hek:


webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -0,0 +1,162 @@
...
+        Enum.map_reduce(tracks_ids, state, fn track_id, state ->
+          Bunch.Access.get_updated_in(state, [:tracks, track_id], &%Track{&1 | enabled?: false})
+        end)
+
+      children =
+        tracks_ids
+        |> Enum.flat_map(&[tee: &1, fake: &1])
+        |> Enum.filter(&Map.has_key?(ctx.children, &1))
+
+      children = [endpoint: peer_pid] ++ children
```
mickel8, line 144: Why do we check if `tee` and `fake` with given `track_id` exist in `ctx.children` but we don't do the same with `{:endpoint, peer_id}`?

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -0,0 +1,162 @@
...
+
+    if endpoint == nil or endpoint.terminating? do
+      {:absent, [], state}
+    else
+      {tracks_ids, state} = pop_in(state, [:endpoints_tracks, peer_pid])
+
+      {tracks, state} =
+        Enum.map_reduce(tracks_ids, state, fn track_id, state ->
+          Bunch.Access.get_updated_in(state, [:tracks, track_id], &%Track{&1 | enabled?: false})
+        end)
```
mickel8, line 137: Couldn't we just remove these Tracks from state instead of disabling them? I think that this is a problem we were talking about i.e. inability to remove old tracks?

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -0,0 +1,162 @@
...
+        tracks_ids
+        |> Enum.flat_map(&[tee: &1, fake: &1])
+        |> Enum.filter(&Map.has_key?(ctx.children, &1))
+
+      children = [endpoint: peer_pid] ++ children
+
+      tracks_msgs =
+        flat_map_children(ctx, fn
+          {:endpoint, id} when id != peer_pid ->
+            [forward: {{:endpoint, id}, {:add_tracks, tracks}}]
```
mickel8, line 149: `:add_tracks` can be confusing here because in fact we don't add anything new. Maybe rename it to `{:restart_ice, new_tracks}` in `EndpointBin`

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -0,0 +1,162 @@
...
+  require Membrane.Logger
+
+  def start_link() do
+    Membrane.Pipeline.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl true
+  def handle_init(_opts) do
+    play(self())
+    {:ok, %{tracks: %{}, endpoints_tracks: %{}}}
```
mickel8, line 15: I am wondering about `endpoints_track_ids`. It's probably not important but while reading it's not obvious that in `endpoint_tracks` there are `ids` not `%Tracks{}`.

webrtc/videoroom/lib/videoroom/ws.ex
```diff
@@ -0,0 +1,62 @@
...
+    {:ok, state}
+  end
+
+  @impl true
+  def websocket_handle({:text, msg}, state) do
+    msg = Poison.decode!(msg)
+
+    msg =
+      case msg["event"] do
+        "start" -> {:new_peer, self()}
```
mickel8, line 26: Is it a good idea to identify peers by WS pid? On browser side we are sending `keep_alives` to keep our WS alive. Maybe it would be better to generate some id and use it instead of WS pid. In such scenario we don't have to send keep alives from browser. If WS terminates browser can create a new one and send its request using id we previously assigned to it. If this sounds better we can create issue for this.

mickel8 (changes requested):

webrtc/videoroom/lib/videoroom/pipeline.ex
mat-hek: An endpoint is spawned at the moment the peer connects, while the others on track start. We need this check because a peer can be removed before some tracks start.

webrtc/videoroom/lib/videoroom/pipeline.ex
mat-hek: Initially, I did so, and I got an error saying the order of tracks changed which is forbidden. So the tracks must stay till the end of the room.

webrtc/videoroom/lib/videoroom/ws.ex
mat-hek: Keeping websocket open is useful, for example to determine that somebody has left, and generally makes interaction with the frontend much easier. Also, that's what everyone does.

webrtc/videoroom/lib/videoroom/ws.ex
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  @impl true
+  def websocket_handle({:text, "keep_alive"}, state) do
+    {:ok, state}
+  end
+
+  @impl true
+  def websocket_handle({:text, msg}, state) do
+    msg = Poison.decode!(msg)
```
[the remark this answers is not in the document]
mat-hek: Why should we ignore any messages?

webrtc/videoroom/lib/videoroom/pipeline.ex
mat-hek: Actually, these are kind of new tracks that override the old ones. I think we can change the name though, but I think the API shouldn't expose that adding/updating tracks entails restarting ICE.

webrtc/videoroom/lib/videoroom/ws.ex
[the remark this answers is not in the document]
mickel8: I meant that this will raise if someone sends something that is not json.

mickel8 (approved):

PR #41, opened 2021-03-12
Videoroom Dockerfile

piotrekzyla:


webrtc/videoroom/Dockerfile
```diff
@@ -0,0 +1,76 @@
+FROM elixir:1.11-alpine AS build
+ENV DEBIAN_FRONTEND=noninteractive
```
mspanc, line 2: Unnecessary if you use alpine

mspanc (approved):
Minor fix.

I hadn't tested this locally but generally looks good to me

webrtc/videoroom/Dockerfile
piotrekzyla: Sure - a leftover from stages, where I've been using `elixir:1.11`, not yet an alpine image.

PR #43, opened 2021-03-15
videoroom frontend refactor

mat-hek:


webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -34,9 +25,16 @@ export class MembraneWebRTC {
   private _channel?: Channel;
   private channelId: string;
 
-  private localStream: MediaStream;
+  private localTracks: Set<MediaStreamTrack> = new Set<MediaStreamTrack>();
   private remoteStreams: Set<MediaStream> = new Set<MediaStream>();
   private connection?: RTCPeerConnection;
+  private rtc_config: RTCConfiguration = {
```
Qizot, line 31: If we don't modify the config it could be `readonly`.

webrtc/videoroom/lib/videoroom_web/templates/room/index.html.eex
```diff
@@ -4,6 +4,5 @@
     </a>
     <h3>Room: <%= @room_id %></h3>
     <div id="videochat" class="VideoChat"> </div>
-    <div id="control" class="Control" />
```
Qizot, line 7: I believe that this element is used to display errors coming from websocket. Could be changed to:
```
<div id="videochat-error" class="VideoChatError />
```
and updated respectively in `ui.ts` file.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -51,16 +49,11 @@ export class MembraneWebRTC {
     this._channel = ch;
   }
 
-  constructor(
-    socket: Socket,
-    localStream: MediaStream,
-    channelId: string,
-    callbacks?: Callbacks
-  ) {
+  constructor(socket: Socket, channelId: string, callbacks?: Callbacks, config?:RTCConfiguration) {
```
Qizot, line 52:
```suggestion
  constructor(socket: Socket, channelId: string, callbacks?: Callbacks, config?: RTCConfiguration) {
```

Qizot (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
mat-hek: We modify it in the constructor, is that still readonly?

webrtc/videoroom/lib/videoroom_web/templates/room/index.html.eex
mat-hek: done

webrtc/videoroom/assets/src/membraneWebRTC.ts
Qizot: From what I learned in typescript playground you can do that. It will be readonly outside of construtor.

Qizot (approved):

PR #45, opened 2021-03-19
Update deps. Adjust configs to API changes in ICE

mickel8:


webrtc/videoroom/config/ice.exs
```diff
@@ -0,0 +1,6 @@
+import Config
```
mickel8, line 1: I am not sure if this should be a separate file. Now we have so many config files

Qizot: I'm pretty sure we should not add config files that are not default to elixir ones (I mean the names). If this is hard coded then it should probably be put in `config/config.exs` but I guess we would like to provide stun servers in runtime.

Qizot: But I have no idea what should be the format for turn servers as they require username and password.

Qizot (changes requested):

webrtc/videoroom/config/ice.exs
mat-hek: why can't we just put that in config.exs or runtime.exs?

Qizot (approved):

PR #46, opened 2021-03-26
Mute audio scenario and vad notifications

mickel8:
* handles vad notifications
* adds ability to mute audio
* associates tracks with streams in `MembraneWebRTC`


webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -65,10 +65,10 @@ export class MembraneWebRTC {
   }
 
   public addTrack = (track: MediaStreamTrack, stream: MediaStream) => {
-    if(this.connection) {
+    if (this.connection) {
       throw new Error("Adding tracks when connection is established is not yet supported");
     }
-    this.localTracks.add(track);
+    this.localTracks.add([track, stream]);
```
Qizot, line 71: Are you sure that it works? I believe that javascript compares that elements by reference so you will never end up with unique values. 

```js
const s = new Set();
s.add([1,2]);
s.add([1,2]);
```
This set will have 2 entries instead of 1 because array references are compared.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -101,12 +101,31 @@ export class MembraneWebRTC {
...
   };
 
+  public muteMicrophone = async () => {
+    this.localTracks.forEach(([track, stream]) => {
+      if (track.kind == "audio") {
+        track.enabled = !track.enabled;
+        this.channel.push("mute-microphone", { data: !track.enabled, mid: this.getSendTrackMid(track) })
+      }
+    });
+    var buttonText = document.getElementById("mute-microphone")?.innerText
```
Qizot, line 111: Don't use `var`. Use `let` if you intend to change this variable reference and `const` otherwise.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -101,12 +101,31 @@ export class MembraneWebRTC {
...
+    if (buttonText == "Mute microphone") {
```
Qizot, line 112: Consider putting those string `Mute microphone` and `Unmute microphone` into some constants in `conts.ts` file or at least in this file.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -101,12 +101,31 @@ export class MembraneWebRTC {
...
+      document.getElementById("mute-microphone")!.innerText = "Unmute microphone";
```
Qizot, line 113: You query the same elements twice. Do it once, query button and then you can check for it's innerText and then change it.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -101,12 +101,31 @@ export class MembraneWebRTC {
...
+    } else if (buttonText == "Unmute microphone") {
+      document.getElementById("mute-microphone")!.innerText = "Mute microphone";
+    }
+  }
+
+  private getSendTrackMid(track: MediaStreamTrack) {
```
Qizot, line 119:
```suggestion
  private getSendTrackMid(track: MediaStreamTrack): <here goes return type> {
```

And either change that function to `findTrackTransceiver` and then get `mid` param or rename it to something more descriptive e.g. `retrieveSenderTrackMid`.

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -33,27 +33,42 @@ defmodule VideoRoomWeb.RoomChannel do
...
 
+  @impl true
   def handle_in("candidate", %{"data" => %{"candidate" => candidate}}, socket) do
     socket
     |> send_to_pipeline({:signal, self(), {:candidate, candidate}})
 
     {:noreply, socket}
   end
 
+  @impl true
```
Qizot, line 52: Isn't putting `@impl true` once enough?

Qizot (changes requested):

webrtc/videoroom/lib/videoroom_web/room_channel.ex
Qizot: And I believe that elixir at its core uses it once: https://github.com/elixir-lang/elixir/blob/e523eaa53d0e16c65b5c330a3963b36721ab037f/lib/elixir/lib/string_io.ex#L191-L202

mat-hek: Using `@impl` in each clause makes it catch wrong arity/typo mistakes in each clause, that's why we tend to do so

mickel8:
> adds ability to mute audio
> associates tracks with streams in MembraneWebRTC

have been done in #51 

> handles vad notifications

has been moved to #56

PR #48, opened 2021-03-30
Videoroom screensharing

### LINKED ISSUE #50: Videoroom Screensharing

Qizot:
Closes #50


webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -101,12 +123,84 @@ export class MembraneWebRTC {
...
+        );
+        if (
+          (event.target as RTCPeerConnection).iceConnectionState === "connected"
+        ) {
+          this.channel.push("connected", {});
+        }
+      };
+      this.fakeStream
+        .getTracks()
+        .forEach((track) => this.connection!.addTrack(track, this.fakeStream));
```
mat-hek, line 203: I wouldn't add any fake streams by default. Instead, we could have `addFakeTrack` (as an alias for `addTrack(fakeTrack, ...)`) and `replaceTrack`, to stay close to the WebRTC API

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -147,19 +241,64 @@ export class MembraneWebRTC {
...
-        this.callbacks.onRemoveTrack?.(event.track, stream);
+
+        // if stream is an active screensharing and has no tracks trigger screensharing end callback
+        if (
+          isScreenSharing &&
+          !hasTracks &&
+          stream === this.remoteScreensharingStream
+        ) {
+          this.callbacks?.onScreensharingEnd?.();
+        }
```
mat-hek, line 265: Can't we just trigger it when removing the screensharing track?

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -1,3 +1,17 @@
+import { MembraneWebRTC } from "./membraneWebRTC";
+
+interface State {
+  isLocalScreensharingActive: boolean;
+  isRemoteScreensharingActive: boolean;
+  screensharingController?: MembraneWebRTC;
```
mat-hek, line 6: MembraneWebRTC should be responsible for taking tracks and sending them, not controlling screen sharing - that's the duty of room.ts

mat-hek (changes requested):
Wrote some comments on how I see it, though let's wait for @bblaszkow06 and discuss other options

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -68,8 +95,32 @@ defmodule VideoRoomWeb.RoomChannel do
...
+    {:noreply, socket}
+  end
+
+  def handle_info({:DOWN, _ref, :process, _monitor, reason}, socket) do
+    push(socket, "error", %{
+      error: "Room stopped working, consider restarting your connection, #{inspect(reason)}"
+    })
+
     {:noreply, socket}
   end
```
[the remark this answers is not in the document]
Qizot: If we kill the channel then we don't know if the pipeline failed or the channel itself and I suppose it is a useful information on frontend.

webrtc/videoroom/assets/src/membraneWebRTC.ts
Qizot: I don't know if I understand you correctly, but it is necessary for a stream to be there when negotiation is being performed. We have to provide fake stream as we don't support dynamic adding/removing of tracks in our pipeline.

webrtc/videoroom/assets/src/membraneWebRTC.ts
Qizot: It does not work like this. Remote screensharing can end in two ways:
* user explicitly stopped it and the information is sent via websocket
* user left the room unexpectedly without explicit exit therefore we have to rely on webrtc `ontrackremove`

When user explicitly stops screensharing the track is not being removed, it just stops sending media. In the second case the stream just ends.

This is why I handle screensharing end in two places.

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -47,6 +47,33 @@ defmodule VideoRoomWeb.RoomChannel do
...
+  #
+  # as peer can get connected several times (after each renegotiation)
+  # respond just to the first one
+  def handle_in("connected", _, socket) do
+    if is_nil(socket.assigns[:connected]) do
+      socket |> send_to_pipeline({:connected, self()})
+    end
+
+    {:noreply, assign(socket, :connected, true)}
+  end
```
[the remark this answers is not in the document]
Qizot: I changed the flow so right now there should not be any RC.

PR #49, opened 2021-03-30
Actions and Swarm Deployment

piotrekzyla:


webrtc/videoroom/config/runtime.exs
```diff
@@ -17,8 +16,8 @@ get_env = fn env, default ->
   end
 end
 
-host = get_env.("HOST", "localhost")
-port = get_env.("PORT", default_port) |> String.to_integer()
+host = get_env.("VIRTUAL_HOST", "localhost")
+port = 4000
```
Qizot, line 20: If we change the port to fixed `4000` we should probably update `README.md` to reflect that.

Qizot (changes requested):

webrtc/videoroom/config/runtime.exs
piotrekzyla: good point!

Qizot (approved):

.github/workflows/webrtc_videoroom_build_on_main.yml
```diff
@@ -0,0 +1,44 @@
+name: WebRTC Videoroom Build on Main
+on:
+  push:
+    branches: 
+      - "main"
```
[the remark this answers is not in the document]
piotrekzyla: fixed

PR #56, opened 2021-04-07
Handle VAD notifications

mickel8:


Qizot (approved):
Seems to be working, I don't know what is the delay though and if it is good enough for switching video streams in the future.

mickel8:
> Seems to be working, I don't know what is the delay though and if it is good enough for switching video streams in the future.

@Qizot it's configurable so we will adjust it in the future

PR #62, opened 2021-04-09
Limit track number

mickel8:


webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -63,7 +73,7 @@ defmodule VideoRoom.Pipeline do
 
   @impl true
   def handle_other({:new_peer, peer_pid, peer_type, ref}, ctx, state) do
-    send(peer_pid, {:new_peer, :ok, ref})
+    send(peer_pid, {:new_peer, {:ok, 10}, ref})
```
mat-hek, line 76: shouldn't be `@max_display_num`?

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -171,17 +213,60 @@ defmodule VideoRoom.Pipeline do
...
+                actions ++
+                  [
+                    {:forward, {{:endpoint, id}, {:disable_track, old_track_id}}},
+                    {:forward, {{:endpoint, id}, {:enable_track, new_track_id}}}
+                  ]
+
+              # send(endpoint_id, {:signal, {:replace_track, old_track_id, new_track_id}})
+              {actions, Map.put(display_managers, id, display_manager)}
+
+            {:error, :no_such_endpoint_id} ->
```
mat-hek, line 257: shouldn't we fail here?

mat-hek (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -43,6 +43,7 @@ export class MembraneWebRTC {
   private channelId: string;
   private socketRefs: string[] = [];
 
+  private max_display_num: number = 1;
```
Qizot, line 46: use camel case

webrtc/videoroom/lib/videoroom/display_manager.ex
```diff
@@ -0,0 +1,149 @@
+defmodule VideoRoom.DisplayManager do
```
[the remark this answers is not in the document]
Qizot: Documentation on what it does would be useful. It is not mentioned either here neither in pipeline. If somebody  new encounters this he would be quite overwhelmed.-

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -316,4 +414,60 @@ defmodule VideoRoom.Pipeline do
...
+              {actions, display_managers}
+          end
+
+        {actions, display_managers}
+    end)
```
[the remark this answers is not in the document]
Qizot: Agree, the pipeline module is huge right now.

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -85,11 +85,24 @@ defmodule VideoRoomWeb.RoomChannel do
     {:noreply, socket}
   end
 
+  @impl true
+  def handle_info({:signal, {:replace_track, old_track_id, new_track_id}}, socket) do
```
Qizot, line 89: I think that you send this information about replacing track for each infividual endpoint (correct me if I'm wrong) so you should probably not broadcast this information (you end up with `n^2` messages instead of `n`). Simple `push` should do the job.

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -85,11 +85,24 @@ defmodule VideoRoomWeb.RoomChannel do
...
+    broadcast!(socket, "replace_track", %{
+      data: %{
+        "type" => "replace_track",
+        "old_track_id" => old_track_id,
```
Qizot, line 93: And if you send some maps that will be then converted to json rather use camel case for them as javascript does not really like snake case.

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -85,11 +85,24 @@ defmodule VideoRoomWeb.RoomChannel do
...
+    {:noreply, socket}
+  end
+
   @impl true
   def handle_info({:new_peer, response, ref}, socket) do
     case response do
-      :ok ->
-        reply(ref, {:ok, %{}})
+      {:ok, max_display_num} ->
+        reply(ref, {:ok, %{max_display_num: max_display_num}})
```
Qizot, line 105: As mentioned before, camel case.

Qizot (changes requested):

Qizot:
Before you get this merged and deployed could you please make sure that everything related to `membrane_ice_plugin` and `ex_dtls` with `membrane_dtls_plugin` is up to date with their master branches so videoroom does not have any memory/cnode leaks?

mickel8:
@Qizot sure. I think that updating ice in videoroom should be done in separate PR but I will make sure to do this before deploying a new version

webrtc/videoroom/lib/videoroom/pipeline.ex
mickel8: I was thinking about it but raising here will kill the whole room

mat-hek: Ok, but can this error ever happen?

mickel8: Acctually no so I removed it entirely

mickel8:
@Qizot @mat-hek could you please mark comment as resolved if something has been fixed? It will be easier for me to see what should still be improved :slightly_smiling_face:

webrtc/videoroom/assets/src/room.ts
```diff
@@ -89,6 +90,8 @@ const setup = async () => {
             removeVideoElement(track, stream);
           }
         },
+        onServerReplaceStream: replaceStream,
```
Qizot, line 93: Do we need `Server` in callback name?

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -42,11 +42,8 @@ export function getRoomId(): string {
   return document.getElementById("room")!.dataset.roomId!;
 }
 
-export function addVideoElement(
-  _: MediaStreamTrack,
-  stream: MediaStream,
-  mute: boolean = false
-) {
+export function addVideoElement(stream: MediaStream, mute: boolean = false) {
+  console.log(stream);
```
Qizot, line 46: Unnecessary log.

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -132,6 +129,20 @@ export function setErrorMessage(
   }
 }
 
+export function replaceStream(oldStreamId: String, newStream: MediaStream) {
+  const grid = document.getElementById("videos-grid");
+  var videoElements = grid?.getElementsByTagName("video")!;
+  for (var i = 0; i < videoElements.length; i++) {
```
Qizot, line 135: Use `let` if variable's reference is meant to be replaced or `const` otherwise.

Qizot: `let` in this case

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -132,6 +129,20 @@ export function setErrorMessage(
   }
 }
 
+export function replaceStream(oldStreamId: String, newStream: MediaStream) {
```
Qizot, line 132:
```suggestion
export function replaceStream(oldStreamId: String, newStream: MediaStream): void {
```

webrtc/videoroom/assets/src/room_ui.ts
Qizot: And more appropriate would be:
```js
const oldVideo = videoElements.find(video => video.id === oldStreamId);
if (oldVideo) ...
```

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -1,12 +1,15 @@
...
 
-  alias Membrane.WebRTC.{EndpointBin, Track}
+  alias Membrane.WebRTC.{EndpointBin, Track, Endpoint}
+  alias VideoRoom.DisplayEngine
 
   require Membrane.Logger
 
   @pipeline_registry VideoRoom.PipelineRegistry
 
+  @max_display_num 3
```
Qizot, line 11: Would be cool if we put that inside runtime config and set the default to 3

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -171,17 +205,25 @@ defmodule VideoRoom.Pipeline do
...
-  def handle_notification({:signal, message}, {:endpoint, peer_pid}, _ctx, state) do
-    send(peer_pid, {:signal, message})
-    {:ok, state}
+  def handle_notification(
+        {:vad, val} = msg,
+        {:endpoint, endpoint_id} = from,
+        _ctx,
+        state
+      ) do
+    Membrane.Logger.info("#{inspect(msg)}, from: #{inspect(from)}")
```
Qizot, line 219: Isn't it rather a debug level information than info?

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -85,11 +85,30 @@ defmodule VideoRoomWeb.RoomChannel do
     {:noreply, socket}
   end
 
+  @impl true
+  def handle_info({:signal, {:replace_track, old_track_id, new_track_id}}, socket) do
+    push(socket, "replaceTrack", %{
+      data: %{
+        "type" => "replaceTrack",
```
Qizot, line 92: do we need this "type" if we don't support any other type in `replaceTrack` event?

Qizot (changes requested):

webrtc/videoroom/assets/src/room.ts
mickel8: The `Server` word is to inform that this callback will be invoked on server indication. In other words there will be situations when server will tell us to replace stream. I think that it's a little clearer in this form but don't insist.

Qizot: I'm wondering if this is necessary here as this callback is exposed to UI so it can react to stream change, it should not care how this happens.

webrtc/videoroom/assets/src/room_ui.ts
mickel8: I can't use `find` on `HTMLCollectionOf<HTMLVideoElement>`

webrtc/videoroom/assets/src/room_ui.ts
mickel8: We didn't add `: void` in other functions' definition. Do we want to put it everywhere?

webrtc/videoroom/lib/videoroom/pipeline.ex
mickel8: Acctually this should be totally removed

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -274,6 +320,19 @@ defmodule VideoRoom.Pipeline do
     []
   end
 
+  defp parse_max_display_num(max_display_num_raw) do
+    case Integer.parse(max_display_num_raw) do
```
Qizot, line 324: This can be probably removed. Just put `String.to_integer()` inside `runtime.exs` and it will crash the application on invalid config value.

webrtc/videoroom/config/runtime.exs
```diff
@@ -4,7 +4,8 @@ import Config
 # turn_servers: "addr:port:username:password:proto"
 config :membrane_videoroom_demo,
   stun_servers: System.get_env("STUN_SERVERS", "64.233.163.127:19302"),
-  turn_servers: System.get_env("TURN_SERVERS", "")
+  turn_servers: System.get_env("TURN_SERVERS", ""),
+  max_display_num: System.get_env("MAX_DISPLAY_NUM", "3")
```
Qizot, line 8:
```suggestion
  max_display_num: System.get_env("MAX_DISPLAY_NUM", "3") |> String.to_integer()
```

At least the config will crash before application startup and we don't have to parse it later.

Qizot (changes requested):

webrtc/videoroom/assets/src/room_ui.ts
Qizot: I guess the answer is yes :)

Qizot:
And one more thing. Don't push force the branch as github then loses track of what has changed since last review and I can't see only the updated parts.

webrtc/videoroom/assets/src/room_ui.ts
Qizot: Go resolve if done

webrtc/videoroom/config/runtime.exs
mickel8: This way we allow for passing non positive numbers. Not sure if we want to do this?

mickel8:
> And one more thing. Don't push force the branch as github then loses track of what has changed since last review and I can't see only the updated parts.

Yeah, I know, sorry but had to do rebase :c

webrtc/videoroom/config/runtime.exs
Qizot: Then we may want to do the checking in the config anyways? There is a lot of logic in those files anyway so putting the parsing in the config file would just simplify them.

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -177,17 +208,19 @@ defmodule VideoRoom.Pipeline do
         end)
 
     spec = %ParentSpec{children: children, links: links}
-    state = update_in(state, [:tracks, track_id], &%Track{&1 | encoding: encoding})
+    endpoint = Endpoint.update_track_encoding(state.endpoints[endpoint_id], track_id, encoding)
+    state = put_in(state.endpoints[endpoint_id], endpoint)
```
mat-hek, line 212:
```suggestion
    state = update_in(state, [:endpoints, endpoint_id], &Endpoint.update_track_encoding(&1, track_id, encoding))
```

webrtc/videoroom/lib/videoroom/display_manager.ex
```diff
@@ -0,0 +1,170 @@
...
+          {nil, nil} ->
+            state = %{state | queued: MapSet.put(state.queued, endpoint_id)}
+            state = %{state | rest: MapSet.delete(state.rest, endpoint_id)}
+            {:ok, state}
+
+          {inactive_endpoint_id, :silence} ->
+            state = %{state | displayed: Map.delete(state.displayed, inactive_endpoint_id)}
+            state = %{state | rest: MapSet.put(state.rest, inactive_endpoint_id)}
+            state = %{state | rest: MapSet.delete(state.rest, endpoint_id)}
+            state = put_in(state.displayed[endpoint_id], :speech)
```
mat-hek, line 70: I think a `move` utility function, that would enable moving endpoint id from one set to another, would make this prettier

webrtc/videoroom/lib/videoroom/display_engine.ex
```diff
@@ -0,0 +1,183 @@
...
+                |> case do
+                  %Track{id: id} -> id
+                end
+
+              new_track_id =
+                Endpoint.get_video_tracks(state.endpoints[new_id])
+                |> List.first()
+                |> case do
+                  %Track{id: id} -> id
+                end
```
mat-hek, line 93: What's the difference between the id returned from DisplayManager and the one taken from Endpoint?

webrtc/videoroom/lib/videoroom/display_engine.ex
```diff
@@ -0,0 +1,183 @@
...
+    {actions, state}
+  end
+
+  @doc """
+  Returns information if video track of endpoint with id `endpoint_id2` should be displayed on endpoint with id
+  `endpoint_id1` i.e. if `endpoint_id1` has free space for video track.
+  """
+  @spec display?(state :: t(), endpoint_id1 :: Endpoint.id(), endpoint_id2 :: Endpoint.id()) ::
+          boolean()
+  def display?(state, endpoint_id1, endpoint_id2),
```
mat-hek, line 115:
```suggestion
  def display?(state, target_endpoint_id, source_endpoint_id),
```

mat-hek (changes requested):

webrtc/videoroom/lib/videoroom/display_engine.ex
mickel8: DisplayManager returns id of Endpoint which video track should be replaced

mat-hek: oh ok, so let's extract 
```elixir
Endpoint.get_video_tracks(state.endpoints[old_id])
|> List.first()
|> case do
  %Track{id: id} -> id
end
```
to a private function

mat-hek: Though I'm wondering if DisplayEngine should rather operate on tracks and streams, not only endpoints, so it could return the track IDs directly

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -26,9 +26,11 @@ interface TrackContext {
 }
 
 interface Callbacks {
-  onAddTrack?: (ctx: TrackContext) => void;
+  onAddTrack?: (ctx: TrackContext, display: boolean) => void;
   onRemoveTrack?: (ctx: TrackContext) => void;
   onConnectionError?: (message: string) => void;
+  onReplaceStream?: (oldStreamId: String, newStream: MediaStream) => void;
+  onDisplayStream?: (stream: MediaStream) => void;
```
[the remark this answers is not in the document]
mickel8: We don't know which tracks should be displayed. We only know which track should be replaced.

webrtc/videoroom/lib/videoroom/display_engine.ex
mickel8: The first version was based on Track ids but it turned out to be a little easier to operate on Endpoint ids. I would leave it for now and change it while limiting audio to the most loud ones

webrtc/videoroom/assets/src/membraneWebRTC.ts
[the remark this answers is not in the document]
mat-hek: Hmm ok, so what I don't like about this API is that:
- it imposes the way that the speakers are replaced, which is quite a frontend responsibility
- it is quite high-level, while we expect MembraneWebRTC to be a low-level, general-purpose utility
- it's stream-based, while the entire API is track-based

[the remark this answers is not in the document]
mickel8: > it imposes the way how the speakers are replaced, which is quite a frontend responsibility
it is quite high-level, while we expect MembraneWebRTC to be a low-level, general-purpose utility

These two callbacks are invoked in response to server notifications. So when server sends notification `{replace_track, track_id1, track_id2}` we invoke `onReplaceStream` as passing track here has no sense. 

I can change it but I don't exactly know what is a preffared way. There is no extra logic in membraneWebRTC.ts. It just invokes callback when notification arrives.

In the first version we had `onServerReplaceStream` which was to indicate that this callback is invoked in resposne to server demand. 

> it's stream-based, while the entire API is track-based

I am not sure why it's track-based as in fact we don't use `MediaStreamTrack` in any callback implementation.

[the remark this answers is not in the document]
mickel8: Also sending notification from server `replace_stream, stream_id1, stream_id2` would be a better idea but I would wait for unifying Track.id with msid etc.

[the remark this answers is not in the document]
mickel8: Maybe let'schange `onReplaceStream` to `onServerReplaceTrack(track1, stream1, track2, stream2)`
and `onDisplayStream` to `onServerDisplayTrack`?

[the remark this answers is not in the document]
mat-hek: > I am not sure why it's track-based as in fact we don't use MediaStreamTrack in any callback implementation.

We have onAdd**Track** and onRemove**Track** and they accept **Track**Context, that wraps said MediaStreamTrack ;)

> we invoke onReplaceStream as passing track here has no sense

That's not that obvious, because what actually happens is that we prune some video _tracks_ and unprune others. That's why I said the API is high-level: it introduces case-specific semantics. In another use case, we may need the pruning but it will have nothing to replacing streams.

I don't insist to change that now, though I'm trying to show my point.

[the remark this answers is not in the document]
mickel8: > We have onAddTrack and onRemoveTrack and they accept TrackContext, that wraps said MediaStreamTrack ;)

Yeah I meant that implementation of `onAddTrack` `onRemoteTrack` etc. ignores tracks so we have track based API but we don't use these tracks anywhere.
 
> I don't insist to change that now, though I'm trying to show my point.
I see. I mean I know this might not look perfect but as long as changing this doesn't require any changes on the server side (i.e. we can say that server side API is ok)  I would leave it.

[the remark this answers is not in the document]
mat-hek: FYI the track-based API comes from WebRTC API, which was stream-based, but they made it track-based. Let's leave it then, but I'd rethink it when we get back to limiting audio.

webrtc/videoroom/lib/videoroom/display_manager.ex
```diff
@@ -0,0 +1,175 @@
...
+      Map.has_key?(state.displayed, endpoint_id) ->
+        if MapSet.size(state.queued) == 0 do
+          state = put_in(state.displayed[endpoint_id], :silence)
+          {:ok, state}
+        else
+          queued_endpoint_id = MapSet.to_list(state.queued) |> List.first()
+          state = %{state | displayed: Map.delete(state.displayed, endpoint_id)}
+          state = %{state | rest: MapSet.put(state.rest, endpoint_id)}
+          state = %{state | queued: MapSet.delete(state.queued, queued_endpoint_id)}
+          state = %{state | displayed: Map.put(state.displayed, queued_endpoint_id, :speech)}
```
mat-hek, line 87: my point was to use swap also here

webrtc/videoroom/lib/videoroom/display_manager.ex
```diff
@@ -0,0 +1,175 @@
...
+          rest_endpoint_id = MapSet.to_list(state.rest) |> List.first()
+
+          if rest_endpoint_id == nil do
+            # there is no alternative to display
+            state = %{state | displayed: Map.delete(state.displayed, endpoint_id)}
+            {:ok, state}
+          else
+            state = %{state | displayed: Map.delete(state.displayed, endpoint_id)}
+            state = %{state | rest: MapSet.delete(state.rest, rest_endpoint_id)}
+            state = %{state | displayed: Map.put(state.displayed, rest_endpoint_id, :silence)}
```
mat-hek, line 128: and here

webrtc/videoroom/lib/videoroom/display_manager.ex
```diff
@@ -0,0 +1,175 @@
...
+            {{:replace, endpoint_id, rest_endpoint_id}, state}
+          end
+        else
+          queued_endpoint_id = MapSet.to_list(state.queued) |> List.first()
+          state = %{state | displayed: Map.delete(state.displayed, endpoint_id)}
+          state = %{state | queued: MapSet.delete(state.queued, queued_endpoint_id)}
+          state = %{state | displayed: Map.put(state.displayed, queued_endpoint_id, :speech)}
```
mat-hek, line 135: and here

mat-hek (commented):
And fix conflicts and we can merge ;)

mickel8:
@mat-hek should we squash these commits?

mat-hek:
@mickel8 it's up to you

webrtc/videoroom/assets/src/membraneWebRTC.ts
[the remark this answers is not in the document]
Qizot: The WebRTC may be track based but what is passed to video element is still a media stream and there is no way to directly add tracks to it without changing passed stream.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -104,14 +109,28 @@ export class MembraneWebRTC {
 
     this.channel.on("offer", this.onOffer);
     this.channel.on("candidate", this.onRemoteCandidate);
+    this.channel.on("replaceTrack", (data: any) => {
+      const oldTrackId = data.data.oldTrackId;
```
Qizot, line 113: are you sure typescript is not complaining about the any type here?

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -104,14 +109,28 @@ export class MembraneWebRTC {
...
+    });
 
     this.channel.on("error", (data: any) => {
       this.callbacks.onConnectionError?.(data.error);
       this.stop();
     });
 
     await phoenix_channel_push_result(this.channel.join());
-    await phoenix_channel_push_result(this.channel.push("start", {}));
+    await phoenix_channel_push_result(this.channel.push("start", {})).then(
```
Qizot, line 131: `await` already returns the result, you could do something like this:
```js
const { maxDisplayNum } = await phoenix_channel_push_result(this.channel.push("start", {}));
this.maxDisplayNum = maxDisplayNum;
```

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -175,16 +194,22 @@ export class MembraneWebRTC {
   private onTrack = () => {
     return (event: RTCTrackEvent) => {
       const [stream] = event.streams;
-      this.remoteStreams.add(stream);
+      const mid = event.transceiver.mid!;
+      const isScreenSharing = mid.includes("SCREEN") || false;
 
-      const isScreenSharing =
-        event.transceiver.mid?.includes("SCREEN") || false;
+      isScreenSharing
```
Qizot, line 200: Use `if` statement here instead of ternary.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -175,16 +194,22 @@ export class MembraneWebRTC {
...
+        ? (this.screensharingStream = stream)
+        : this.remoteStreams.add(stream);
+      this.midToStream.set(mid, stream);
 
       stream.onremovetrack = (event) => {
         const hasTracks = stream.getTracks().length > 0;
 
         if (!hasTracks) {
-          this.remoteStreams.delete(stream);
+          isScreenSharing
```
Qizot, line 209: Avoid ternary here again.

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -43,27 +43,42 @@ export function getRoomId(): string {
...
+
+  video.id = stream.id;
   video.srcObject = stream;
   video.autoplay = true;
   video.playsInline = true;
   video.muted = mute;
 }
 
-export function removeVideoElement(_: MediaStreamTrack, stream: MediaStream) {
+export function addAudioElement(stream: MediaStream): void {
```
Qizot, line 66: Are you sure that we aren't playing audio twice? If we add stream to video element and it contains audio track it will just play it. If we create another audio element and put there the stream it will play audio again.

webrtc/videoroom/config/runtime.exs
```diff
@@ -1,10 +1,76 @@
 import Config
 
+defmodule ConfigParser do
```
Qizot, line 3: I guess it could land in another file in `config/` directory?

Qizot (changes requested):

mickel8:
Moved to #74

webrtc/videoroom/assets/src/room_ui.ts
mickel8: Thought about it. I didn't notice any problems so far

webrtc/videoroom/assets/src/membraneWebRTC.ts
mickel8: I don't see any warnings

webrtc/videoroom/config/runtime.exs
mickel8: We would have to put it in another `.exs` file and require it here. We only use it here so I don't think it's worth

webrtc/videoroom/assets/src/room_ui.ts
Qizot: The audio may be in sync and if 2 of them are playing we are introducing some overhead for the browser.

webrtc/videoroom/config/runtime.exs
Qizot: Just thought it would look more clean.

webrtc/videoroom/assets/src/room_ui.ts
mickel8: True that, I would recall here if we notice any problems with browser performance or while limiting audio tracks to those that are the loudest. It has been working pretty well so far (I mean stable).

PR #63, opened 2021-04-09
Updated readme with how to run demo with docker

Qizot:


webrtc/videoroom/README.md
```diff
@@ -52,6 +52,36 @@ mix phx.server
 
 Then, go to <http://localhost:4000/>.
 
+### Run with docker
+
+Videoroom demo provides a `Dockerfile` that you can use to build run videoroom application yourself without any additional setup and dependencies.
```
mickel8, line 57:
```suggestion
Videoroom demo provides a `Dockerfile` that you can use to run videoroom application yourself without any additional setup and dependencies.
```

webrtc/videoroom/README.md
```diff
@@ -52,6 +52,36 @@ mix phx.server
...
+First copy `.env.example` with necessary environmental variables:
+
+```bash
+cp .env.example .env
+```
+
+**IMPORTANT**
+If you intend to use TLS remember that setting paths in `.env` file is not enough.
+Those paths will be used inside docker container therefore besides setting env variables you will need to mount those paths
+to docker container yourself.
```
mickel8, line 70:
```suggestion
to docker container on your own.
```

webrtc/videoroom/.env.example
```diff
@@ -0,0 +1,7 @@
+VIRTUAL_HOST=localhost
```
mickel8, line 1: Do we want it to be named `.env.example`? Why not `.env`? It's demo so I would put it as `.env` and remove instruction to copy `.env.example` to `.env` in README.

webrtc/videoroom/README.md
```diff
@@ -52,6 +52,36 @@ mix phx.server
 
 Then, go to <http://localhost:4000/>.
 
+### Run with docker
```
mickel8, line 55: Maybe add following sections
## Usage
### Environment Variables
Available runtime environmental variables:
```
VIRTUAL_HOST={host passed to the endpoint config, defaults to "localhost" on non-production environments}
USE_TLS={"true" or "false", if set to "true" then https will be used and certificate paths will be required}
KEY_FILE_PATH={path to certificate key file, used when "USE_TLS" is set to true}
CERT_FILE_PATH={path to certificate file, used when "USE_TLS" is set to true}
STUN_SERVERS={list of stun servers separated by ",", defaults to a single server "stun1.l.google.com:19302"}
TURN_SERVERS={list of turn servers separated by ",", defaults to ""}
```
### Run manually

Default path for certificate files for non-production environment is `priv/certs/`.
If you want to run TLS locally you can generate certificates
as described in the [signaling server readme](https://github.com/membraneframework/membrane_demo/tree/master/webrtc/simple#https). 
First install all dependencies:
```
mix deps.get
npm ci --prefix=assets
```
In order to run, type:
```
mix phx.server 
```

Then, go to <http://localhost:4000/>.

### Run with docker
...

webrtc/videoroom/README.md
```diff
@@ -52,6 +52,36 @@ mix phx.server
...
+docker run -p 4000:4000 --env-file .env membraneframework/demo_webrtc_videoroom:latest
+```
+
+Or build and run docker image from source:
+```bash
+docker build  -t membrane_videoroom .
+docker run -p 4000:4000 --env-file .env membrane_videoroom 
+```
+
+Application is once again available at <http://localhost:4000/>.
```
mickel8, line 83: I would write `Then go to` as above. Most people will go directly to `Run with docker` I think.

mickel8 (changes requested):

mickel8 (approved):

PR #72, opened 2021-04-21
Participants labeling

### LINKED ISSUE #70: VideoRoom track's owner identification
# Participant identification

Jitsi just like Membrane support multiple peers on a single shared peer connection.

There is a problem of identifying each peer based on the incoming tracks. The SFU needs to
keep track of its participants (things like display name ect.) and pass that information 
to the client when needed.

## How Jitsi does that?

The low-level code that is responsible for identifying incoming tracks with participants is available here: 

https://github.com/jitsi/lib-jitsi-meet/blob/baa78aca40541b87b44d5218bcdef2a7be5fee59/modules/RTC/TraceablePeerConnection.js#L813

Jitsi listens on remote tracks being added. Then it does manual search on the `remoteDescription.sdp` string to obtain necessary information.

First it checks the transceiver for `mid` field, if not found then if falls back to searching for `msid` that equals to the current stream id. 
(It all depends on what plan the browser is using but as far as I know Unifed Plan is now a standard).
Then the sdp is filtered for lines that include `mid` or `msid` fields that has been found in previous step.
In resulting lines  we filter out received lines to get corresponding `ssrc`.

The main goal here is to obtain corresponding `ssrc` of incoming track. Then a request is sent 
to the server with given `ssrc` to return track's participant id. And further down the line if participant 
is not in client's cache, again a request is sent to obtain participant's information.

## How should we do that?

What fields can we use to query information about given participant? Each participant will see the same `mid` values,
from what I recall `msid` will differ for each participant same goes with `ssrc` and when querying the server it will
have to identify the caller to get the proper mapping of outgoing streams (with `mid` it can do a really quick check).

And how should we get participants information, after getting a new track by sending query request or the server itself should first send information about incoming participant?

Qizot:
Closes #70 #71 

For participant identification and labeling I've used `mid -> participant` mapping. When user gets an offer from the server he will also receive a list of participants where each consists of `displayName` and `mids` which is a list of `mid` values for given participant.

Still to do:
- [x] get display name from user when he enters the room


webrtc/videoroom/lib/videoroom/participant_mapper.ex
```diff
@@ -0,0 +1,58 @@
+defmodule VideoRoom.ParticipantMapper do
```
mickel8, line 1: I am wondering about creating GenServer for storing association between `display_name` and `mids`. In fact there is the module [`Membrane.WebRTC.Endpoint`](https://github.com/membraneframework/membrane_webrtc_plugin/blob/master/lib/membrane_webrtc_plugin/endpoint.ex) which aggregates Endpoint's inbound tracks. It could be better to move `display_name` there. We could even create there something like `user_data` to make it more generic.

`Membrane.WebRTC.Endpoint` is integrated with videoroom on branch [`track-filter`](https://github.com/membraneframework/membrane_demo/blob/track-filter/webrtc/videoroom/lib/videoroom/pipeline.ex#L57)

webrtc/videoroom/assets/src/room.ts
```diff
@@ -98,12 +107,13 @@ const setup = async () => {
...
       webrtc.addTrack(track, localStream);
-      addVideoElement(track, localStream, true);
+      addVideoElement(track, localStream, "Me", true);
     });
 
     setupRoomUI({
       state: {
-        onLocalScreensharingStart: () => startLocalScreensharing(socket),
+        onLocalScreensharingStart: () =>
+          startLocalScreensharing(socket, displayName),
```
mickel8, line 116: This should be sth like `My screeensharing` instead of `displayName`

webrtc/videoroom/lib/videoroom/participant_mapper.ex
Qizot: The reason behind putting that mapper in other genserver was to reduce pipeline's state manipulation complexity as adding another function modifying the state would just make everything worse. 

I wouldn't go for putting things such `display_name` in `Membrane.WebRTC.Endpoint` as it is low-level abstraction and it probably should not be aware of participant specific information. It doesn't event know about `peer_pid` which is only tracked in pipeline by monitoring the pid.

mickel8: I don't still think we need a separate proccess here. It looks more like we use GenServer for code organization. We will also distribute state among different repositories. The idea is not to store display_name directly in an Endpoint but something like `user_data` of type `any` so the API would look like this

```elixir
Endpoint.get_user_data/1
Endpoint.update_user_data/1
```

In such scenario we can associate any arbitrary data with the Endpoint and don't need to care about cleaning it when the Endpoint disconnects from server.

Qizot: Ok, seems reasonable but I would opt for different name as the name `user` does not fit this level of abstraction.

Qizot: And yet I find another problem with this approach. When we decide to keep such data in `Endpoint` we will end up with having to query each endpoint for its metadata. With current approach we keep everything in a single place and we are done with a single request to the gen server.

mickel8: > Ok, seems reasonable but I would opt for different name as the name user does not fit this level of abstraction.

`user_data` was an example we can name it `data` `ctx` `state` etc.

> And yet I find another problem with this approach. When we decide to keep such data in Endpoint we will end up with having to query each endpoint for its metadata. With current approach we keep everything in a single place and we are done within a single request to the gen server.

But it's a simple Enum.map operation, isn't it? We don't need to think about optimizing it at this stage

Qizot: > But it's a simple Enum.map operation, isn't it? We don't need to think about optimizing it at this stage
How is that a simple enum? Don't we have to query each endpiont which will end up in sending and receiving messages?

mickel8: I mean `Endpoin.t()` not `EndpointBin`. If we had `data` in [`Endpoint.t()`](https://github.com/membraneframework/membrane_webrtc_plugin/blob/d1286e593e00f78174ed056d73fda0df33afcb9d/lib/membrane_webrtc_plugin/endpoint.ex#L10-L14) we would do 
```elixir
Enum.map(endpoints, fn endpoint -> Endpoint.get_data(endpoint) end)
```

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -222,17 +226,33 @@ defmodule VideoRoom.Pipeline do
...
   def handle_notification({:signal, message}, {:endpoint, peer_pid}, _ctx, state) do
     send(peer_pid, {:signal, message})
     {:ok, state}
   end
 
+  def handle_notification({:vad, val}, {:endpoint, endpoint_id}, _ctx, state) do
+    display_engine = state.display_engine
+    {actions, display_engine} = DisplayEngine.vad_notification(display_engine, val, endpoint_id)
+    {{:ok, actions}, %{state | display_engine: display_engine}}
+  end
```
mickel8, line 254: This has been already added on master. Is it a fix after reabse? :thinking:

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -80,8 +83,12 @@ defmodule VideoRoomWeb.RoomChannel do
     {:noreply, socket}
   end
 
-  def handle_info({:signal, {:sdp_offer, sdp}}, socket) do
-    push(socket, "offer", %{data: %{"type" => "offer", "sdp" => sdp}})
+  def handle_info({:signal, {:sdp_offer, sdp}, participants}, socket) do
+    participants =
+      participants
+      |> Enum.map(&%{"displayName" => &1.display_name, "mids" => &1.mids})
```
mickel8, line 89:
```suggestion
participants = Enum.map(participants, &%{"displayName" => &1.display_name, "mids" => &1.mids})
```

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -84,7 +84,11 @@ defmodule VideoRoom.Pipeline do
       Process.monitor(peer_pid)
 
       tracks = new_tracks(peer_type)
-      endpoint = Endpoint.new(peer_pid, peer_type, tracks)
+
+      endpoint =
+        Endpoint.new(peer_pid, peer_type, tracks)
+        |> Endpoint.put_context(%{display_name: display_name})
```
mickel8, line 90:
```suggestion
endpoint = Endpoint.new(peer_pid, peer_type, tracks, %{display_name: display_name})
```

mickel8 (changes requested):

webrtc/videoroom/lib/videoroom/pipeline.ex
Qizot: Merge probably moved some lines in a way that this looks like something has been inserted.

mickel8 (approved):

PR #74, opened 2021-04-23
Limit track number

mickel8:
Moved from #62


webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -29,6 +29,8 @@ interface Callbacks {
   onAddTrack?: (ctx: TrackContext) => void;
   onRemoveTrack?: (ctx: TrackContext) => void;
   onConnectionError?: (message: string) => void;
+  onReplaceStream?: (oldStreamId: String, newStream: MediaStream) => void;
+  onDisplayStream?: (stream: MediaStream) => void;
```
Qizot, line 33: I think that this module is not so intuitive when it comes to those callbacks. We use `onDisplayStream` to tell frontend to display the stream and with `onAddTrack` we just create audio elements. It's counter intuitive that we don't create video elements but audio elements are fine.

Qizot (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
mickel8: I am opened for suggestion but we have decided to leave it for now here https://github.com/membraneframework/membrane_demo/pull/62#discussion_r617676673

mickel8: Also audio elements are fine as we want to play all of them while we don't want to display each incoming video track

mat-hek (approved):

PR #77, opened 2021-04-27
msid in tracks

FelonEkonom:
Adjusting code to related changes from membraneframework/membrane_webrtc_plugin#8


webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -13,7 +13,7 @@ const phoenix_channel_push_result = async (push: Push): Promise<any> => {
 
 interface Participant {
   displayName: string;
-  mids: string[];
+  ids: string[];
```
Qizot, line 16: Probably should be trackIds. `ids` does not tell much what it is.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -226,16 +226,15 @@ export class MembraneWebRTC {
   private onTrack = () => {
     return (event: RTCTrackEvent) => {
       const [stream] = event.streams;
-      const mid = event.transceiver.mid!;
-      const isScreenSharing = mid.includes("SCREEN") || false;
+      const id = event.track.id;
+      const isScreenSharing = id.includes("SCREEN") || false;
```
Qizot, line 230: Have you checked that screensharing works? I tried to wrap my head around this pr and one in webrtc plugin and got lost easily...

Qizot (changes requested):

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -169,6 +158,49 @@ defmodule VideoRoom.Pipeline do
...
+
+    state = %{state | display_engine: %DisplayEngine{state.display_engine | endpoints: endpoints}}
+
+    mid_to_msid =
+      Map.get(state, :mid_to_msid, %{})
+      |> Map.merge(mid_to_msid)
+
+    state =
+      state
+      |> Map.put(:mid_to_msid, mid_to_msid)
```
mickel8, line 179: Why do we store this in state? I think we don't need these lines

mickel8 (changes requested):

mickel8:
Outdated

PR #79, opened 2021-04-29
Added handling of denied permissions

Qizot:
In this PR it is possible to join room without one of permissions (microphone/camera) but something seems to be broken when trying to join without both of them. Something must be blocking the flow somewhere as nothing happends for a while and then toilet overflow happens almost imediately for other peer that in this case is sending media.


mickel8:
Done in #107

PR #81, opened 2021-04-30
Fix duplicated audio. Fix replacing video tiles

### LINKED ISSUE #78: [Videoroom] Instead of showing up to 4 users, Videoroom shows all users and freezes most quiet ones
It seems, that instead of hiding all, except 4 loudest participants, Videoroom shows them, but with freezed video. When freezed person starts talking, it's video starts to move and another person is freezed

### LINKED ISSUE #80: Fix duplicated audio

mickel8:
closes #78 
closes #80

### DIFF
--- a/webrtc/videoroom/assets/src/membraneWebRTC.ts
+++ b/webrtc/videoroom/assets/src/membraneWebRTC.ts
@@ -127,8 +127,8 @@ export class MembraneWebRTC {
       const newTrackId = data.data.newTrackId;
       const oldStream = this.midToStream.get(oldTrackId)!;
       const newStream = this.midToStream.get(newTrackId)!;
-      const oldTrack = oldStream.getTrackById(oldTrackId)!;
-      const newTrack = newStream.getTrackById(newTrackId)!;
+      const oldTrack = oldStream.getVideoTracks()[0];
+      const newTrack = newStream.getVideoTracks()[0];
       const oldParticipant = this.participants.find(({ mids }) =>
         mids.includes(oldTrackId)
       );
@@ -141,8 +141,8 @@ export class MembraneWebRTC {
         label: oldParticipant?.displayName ?? "",
         isScreenSharing:
           oldParticipant?.mids
-            .find((mid) => mid === oldTrackId)!
-            .includes("SCREEN") || false,
+            .find((mid) => mid === oldTrackId)
+            ?.includes("SCREEN") || false,
       };
       const newCtx = {
         track: newTrack,
@@ -150,8 +150,8 @@ export class MembraneWebRTC {
         label: newParticipant?.displayName ?? "",
         isScreenSharing:
           newParticipant?.mids
-            .find((mid) => mid === newTrackId)!
-            .includes("SCREEN") || false,
+            .find((mid) => mid === newTrackId)
+            ?.includes("SCREEN") || false,
       };
       this.callbacks.onHideTrack?.(oldCtx);
       this.callbacks.onDisplayTrack?.(newCtx);
@@ -159,7 +159,7 @@ export class MembraneWebRTC {
     this.channel.on("displayTrack", (data: any) => {
       const trackId = data.data.trackId;
       const stream = this.midToStream.get(trackId)!;
-      const track = stream.getTrackById(trackId)!;
+      const track = stream.getVideoTracks()[0];
       const participant = this.participants.find(({ mids }) =>
         mids.includes(trackId)
       );
@@ -169,7 +169,9 @@ export class MembraneWebRTC {
         stream,
         label: participant?.displayName ?? "",
         isScreenSharing:
-          participant?.mids.find(trackId)!.includes("SCREEN") || false,
+          participant?.mids
+            .find((mid) => mid === trackId)
+            ?.includes("SCREEN") || false,
       });
     });
 
--- a/webrtc/videoroom/assets/src/membraneWebRTC.ts
+++ b/webrtc/videoroom/assets/src/membraneWebRTC.ts
@@ -36,12 +36,8 @@ interface Callbacks {
   onAddTrack?: (ctx: TrackContext) => void;
   onRemoveTrack?: (ctx: TrackContext) => void;
   onConnectionError?: (message: string) => void;
-  onReplaceStream?: (
-    oldStream: MediaStream,
-    newStream: MediaStream,
-    newLabel: string
-  ) => void;
-  onDisplayStream?: (stream: MediaStream, label: string) => void;
+  onDisplayTrack?: (ctx: TrackContext) => void;
+  onHideTrack?: (ctx: TrackContext) => void;
 }
 
 interface MembraneWebRTCConfig {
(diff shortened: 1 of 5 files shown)


webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -129,26 +125,52 @@ export class MembraneWebRTC {
...
+      const newParticipant = this.participants.find(({ mids }) =>
+        mids.includes(newTrackId)
       );
+      const oldCtx = {
+        track: oldTrack,
+        stream: oldStream,
+        label: oldParticipant?.displayName ?? "",
+        isScreenSharing:
+          oldParticipant?.mids
+            .find((mid) => mid === oldTrackId)!
```
Qizot, line 144: I'm a bit confused right now. Is `mid` the same as `trackId`? I believe that the browser itself sets the track id and what we actually send from backend is just a `mid`. If so, then doing the `oldStream.getTrackById` is just wrong as there will probably be no such track (mid != id).

Besides that, you don't have to use bang operator here:
```
oldParticipants?.mids.find((mid) => mid === oldTrackId)?.includes("SCREEN") || false
```
This should be enough but again I don't know if we are using the `mid` and `trackId` correctly.

webrtc/videoroom/assets/src/room.ts
```diff
@@ -101,19 +100,22 @@ const setup = async () => {
           if (isScreenSharing) {
             setScreensharing(stream, displayName || "", "My screensharing");
           } else {
-            addAudioElement(stream);
+            addVideoElement(stream, displayName || "", false);
```
Qizot, line 103: ahhh, I think that I forgot to clean things up when introducing participat labeling. `getDisplayNamOrRedirect` no longer redirect anywhere, it just takes display name from query params and cleans up the url, can you please change the function name and then replace `displayName || ""` with `displayName`?

Qizot (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
mickel8: `oldTrackId` is `mid` sent by our server so `isScreenSharing` will be set correctly. 

Regaring `oldStream.getTrackById` you are right - this won't work on Firefox but it should work on Chrome as Track.id on chrome is its mid. I forgot about this

Qizot: I don't know about the last thing with chrome setting `trackId` to its `mid`. I specifically remember that I used mids instead of track ids on chrome due to it being auto generated where mid was the same as on the server side.

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -256,24 +278,22 @@ export class MembraneWebRTC {
         });
       };
 
+      const label =
+        this.participants.find((p) => p.mids.includes(mid))?.displayName || "";
       this.callbacks.onAddTrack?.({
```
Qizot, line 283: I think that we have a bug in this place. If we receive two different tracks within the same stream `onAddTrack` will be called twice and frontend will create two audio elements (once for incoming audio track and another for incoming video track).  We are probably mixing too much webrtc logic with business logic, once we are operate on raw tracks and the other time we interpret them as something different. `onAddTrack` should probably stay as it is and frontend should watch out if it is duplicating some video/audio elements.

Qizot (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
mickel8: We no longer create AudioElements. Here we only notify about new incoming track. Frontend will create one VideoElement per stream.

webrtc/videoroom/assets/src/membraneWebRTC.ts
mickel8: You are of course right. This won't work neither in Chrome nor in Firefox :man_facepalming:

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -45,27 +45,35 @@ export function getRoomId(): string {
...
 ): void {
-  const id = elementId(stream, "video");
-  let video = document.getElementById(id) as HTMLVideoElement;
-
-  if (!video) {
-    video = setupVideoFeed(stream, label);
+  const videoId = elementId(stream.id, "video");
+  const audioId = elementId(stream.id, "audio");
+  let video = document.getElementById(videoId) as HTMLVideoElement;
+  let audio = document.getElementById(audioId) as HTMLAudioElement;
```
mat-hek, line 60: do we still use audio elements?

mickel8: I have introduced an audio element in the last two commits. It's needed to hear video tile that has not been displayed yet. Now we have structure as follows:
videoElement - to display video
audioElement - to play sound

If you ask me if it won't  get out of sync - I don't know but first tests are positive.

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -45,27 +45,35 @@ export function getRoomId(): string {
...
 
-  video.id = id;
+  video.id = videoId;
   video.srcObject = stream;
   video.autoplay = true;
   video.playsInline = true;
-  video.muted = mute;
+  video.muted = true;
+
+  audio.id = audioId;
```
Qizot, line 73: Aren't we playing audio twice here? Once in video element and the other time in audio element? We probably don't need audio element anymore

Qizot (changes requested):

webrtc/videoroom/assets/src/room_ui.ts
mickel8: Audio element is needed to hear a person that has not been displayed yet. In other case we have to wait for its video tile. This way we lose about 1 sec of speech. We also always mute video element to not play audio twice

mickel8: Btw. version without audio element is available in this commit 7a4dfc126592b9a21d23454e48bb43b0c696b0b7 if you wanted to test it

webrtc/videoroom/assets/src/room_ui.ts
mat-hek: Ok, any reason why not use video only and show/hide it when needed?

mickel8: Damn, I tested the version with videoElement only but the audio from videoElement is played only when videoElement also receives video frames so it can sync them and play together. This was the reason to introduce audioElements.

mat-hek: Ok, so let's leave it for now. Possibly we can try using  https://stackoverflow.com/questions/6433900/syncing-html5-video-with-audio-playback

mat-hek (approved):

PR #84, opened 2021-05-07
Show participants list

FelonEkonom:


webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -76,6 +76,17 @@ export function addVideoElement(
...
 }
 
+export function setParticipantsNamesList(
+  participantsNames: Array<string>
+): void {
+  const participantsNamesList = document.getElementById(
+    "participants-names-list"
+  ) as HTMLDivElement;
+
+  participantsNamesList.innerHTML =
```
Qizot, line 86: Use `innerText`. I don't know though if list of participants should be displayed in a row instead of column.

Qizot (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -38,6 +38,7 @@ interface Callbacks {
   onConnectionError?: (message: string) => void;
   onDisplayTrack?: (ctx: TrackContext) => void;
   onHideTrack?: (ctx: TrackContext) => void;
+  onOfferDataArrival?: (data: OfferData) => void;
```
mat-hek, line 41:
```suggestion
  onOfferData?: (data: OfferData) => void;
```

mat-hek (changes requested):

mickel8 (changes requested):
I would consider changing two things:
1. Put each participant in a separate row so that we would have
```
Participants:
participant1
participant2
participant3
...
```
2. Accent somehow this list e.g. by using **bold** font for **Participants** and a little larger size

FelonEkonom:
> I would consider changing two things:
> 
>     1. Put each participant in a separate row so that we would have
> 
> 
> ```
> Participants:
> participant1
> participant2
> participant3
> ...
> ```
> 
>     1. Accent somehow this list e.g. by using **bold** font for **Participants** and a little larger size

I think, that maybe it is better idea, to wait with changing look of this list to the moment, when UI developer will come to this project, because I think, that there is very big chance, that we will have to change it then, one way or another

mickel8 (approved):

PR #98, opened 2021-05-11
Add placeholder in place of turned off video and info about muted audio

FelonEkonom:


webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -13,6 +13,8 @@ const phoenix_channel_push_result = async (push: Push): Promise<any> => {
 
 interface Participant {
   displayName: string;
+  mutedAudio: boolean;
+  turnedOffVideo: boolean;
```
mickel8, line 17: I would name it in the same way so maybe `mutedVideo`?

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -44,6 +48,8 @@ interface Callbacks {
   onDisplayStream?: (stream: MediaStream, label: string) => void;
   onDisplayTrack?: (ctx: TrackContext) => void;
   onHideTrack?: (ctx: TrackContext) => void;
+  onToggleVideo?: (streamId: string) => void;
+  onToggleAudio?: (streamId: string) => void;
```
mickel8, line 52: This can be missleading as it can indicate that **we** should toggle video or audio. Maybe `onParticipantToggleAudio` and `onParticipantToggleVideo`

In the future we would have
```ts
onParticipantToggleAudio?: (participantId: string) => void;
onParticipantToggleVideo?: (participantId: string) => void;
```

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -200,6 +220,14 @@ export class MembraneWebRTC {
     }
   };
 
+  public toggleVideo = () => {
+    this.channel.push("toggleVideo", {});
```
mickel8, line 224: `toggledVideo`?

So server will receive info `toggledVideo, peer_pid` and send messages `participantToggledVideo, mid`

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -45,24 +45,41 @@ export function getRoomId(): string {
...
 }
 
 export function addVideoElement(
   stream: MediaStream,
   label: string,
-  muted: boolean = false,
-  isLocalVideo: boolean = false
+  isLocalVideo: boolean = false,
+  mutedAudio: boolean = false,
+  turnedOffVideo: boolean = false,
```
mickel8, line 60: `mutedVideo`?

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -45,24 +45,41 @@ export function getRoomId(): string {
...
+  showMutedAudioIcon: boolean = false
```
mickel8, line 61: Are there any situation when we don't want to show muted audio icon?

mickel8 (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
FelonEkonom: Having varialbe named `mutedVideo` in place, where you want to mute audio in video element (`video.muted = true;`), IMO would be confusing

webrtc/videoroom/assets/src/room_ui.ts
FelonEkonom: User own tile has always muted audio, but we want to show muted audio icon there, only if somebody has truned off his microphone

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -116,6 +116,40 @@ defmodule VideoRoom.Pipeline do
...
+        &Map.update!(&1, peer_pid, fn endpoint ->
+          ctx =
+            case toggle_media do
+              :toggle_video -> %{endpoint.ctx | turned_off_video: !endpoint.ctx.turned_off_video}
+              :toggle_audio -> %{endpoint.ctx | muted_audio: !endpoint.ctx.muted_audio}
+            end
+
+          %Endpoint{endpoint | ctx: ctx}
+        end)
+      )
```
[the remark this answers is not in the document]
FelonEkonom: In this place, `endpoint` is struct `%Endpoint{}`, which does not implement `Access` behaviour, so I cannot write sth like `update_in(endpoint, ...)`

mickel8 (approved):

PR #105, opened 2021-05-18
Fix removing non existing endpoint from DisplayManager. Improve clean up in videoroom

mickel8:


webrtc/videoroom/lib/videoroom/display_manager.ex
```diff
@@ -37,7 +37,7 @@ defmodule VideoRoom.DisplayManager do
   end
 
   @doc """
-  Updates voice activity status of endpoint with id `endpoint_id`.
+  Updates voice activity status of endpoint with id `endpoint_id`. Raises if endpoint doesn't exist.
 
   Returns `{:replace, old_endpoint_id, new_endpoint_id}` if video track of `new_endpoint_id` should be replaced by
```
andpodob, line 42: It sound like returning `{:replace, old_endpoint_id, new_endpoint_id}` causes replacing track of `new_endpoint_id` with `old_endpoint_id` track. That's a bit nonintuitive.

webrtc/videoroom/lib/videoroom/display_manager.ex
```diff
@@ -138,7 +138,8 @@ defmodule VideoRoom.DisplayManager do
         {:ok, state}
 
       true ->
-        raise("No such endpoint id #{inspect(endpoint_id)}")
+        Membrane.Logger.debug("No such endpoint id #{inspect(endpoint_id)}")
+        {:ok, state}
```
andpodob, line 142: Maybe it would be good idea to change `raise` to `debug log` in update function as well? Is that reasonable to crash whole room because we cant update status of peer that actually does not exist?

andpodob, line 142: Maybe it would be good idea to change raise to debug log in update/3 function as well. Is that reasonable to crash whole room because we can't update status of peer that actually does not exist?

andpodob (changes requested):

webrtc/videoroom/lib/videoroom/display_manager.ex
mickel8: This was the first version but in theory it can't happen so we decided to raise. I would leave it for now

webrtc/videoroom/lib/videoroom/display_manager.ex
mickel8: Yes but it's because we operate on Endpoints both in DisplayEngine and DisplayManager. I am opened for suggestions but I would do it in a separate issue

webrtc/videoroom/lib/videoroom/display_manager.ex
andpodob: ok

andpodob (approved):

PR #107, opened 2021-05-20
Join without camera or video

FelonEkonom:


webrtc/videoroom/assets/src/room.ts
```diff
@@ -91,12 +101,36 @@ const setup = async () => {
...
+      console.error("Couldn't get camera permission:", error);
+    }
+
     const webrtc = new MembraneWebRTC(socket, getRoomId(), {
-      displayName,
+      signalingOptions: {
+        displayName,
+        relayAudio: localAudioStream !== null,
+        relayVideo: localVideoStream !== null,
+      },
```
mickel8, line 128: Maybe ParticipantConfig would be more meaningful?

webrtc/videoroom/lib/videoroom/pipeline.ex
```diff
@@ -223,18 +246,24 @@ defmodule VideoRoom.Pipeline do
...
+            muted_audio: ctx.muted_audio,
+            muted_video: ctx.muted_video,
+            mids: Map.keys(tracks)
+          }
+        end)
+
+      user_id = endpoint.ctx.participant_id
+
+      send(peer_pid, {:signal, message, participants, user_id})
+    end
```
mickel8, line 265: Why do we use `with` here? Also function in Enum.map can use get_participant_data

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -97,19 +102,44 @@ defmodule VideoRoomWeb.RoomChannel do
...
+      Enum.map(
+        participants,
+        &%{
+          "id" => &1.id,
+          "displayName" => &1.display_name,
+          "mids" => &1.mids,
+          "mutedAudio" => &1.muted_audio,
+          "mutedVideo" => &1.muted_video
+        }
+      )
```
mickel8, line 139: This is duplicated with lines 106-116 in `handle_info({:signal, {:sdp_offer, sdp}...` Also we have `get_participant_data` function that seems to prepare data to send but here we do almost the same thing by replacing keys to strings. Shouldn't we do this in get_participant_data?

mickel8 (changes requested):

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -48,40 +57,51 @@ interface Callbacks {
...
+    Participant
+  >();
   private readonly rtcConfig: RTCConfiguration = {
     iceServers: [
       {
         urls: "stun:stun.l.google.com:19302",
       },
     ],
   };
+  private userId?: string;
```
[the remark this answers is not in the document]
FelonEkonom: `userId` is equal to value of field `id` in `Participant` representing the recipient of the message

webrtc/videoroom/lib/videoroom_web/room_channel.ex
FelonEkonom: I think, that making such a preparation is dependency, that we don't want to have in pipeline.ex, where get_participants_data is placed, so I extracted functionality of this piece of code to separate, private function and made it more generic

webrtc/videoroom/lib/videoroom/pipeline.ex
FelonEkonom: This `with` is necessary to avoid raise condition when many peers are leaving room at this same time, e.g. when somebody with active screensharing leaves room

webrtc/videoroom/assets/src/membraneWebRTC.ts
```diff
@@ -55,8 +55,8 @@ interface Callbacks {
     newLabel: string
   ) => void;
   onDisplayStream?: (stream: MediaStream, label: string) => void;
```
mickel8, line 57: It seems we messed something up in previous PR's. We no longer use `onReplaceStream` and `onDisplayStream`. Let's remove it

webrtc/videoroom/lib/videoroom/display_engine.ex
```diff
@@ -97,7 +97,11 @@ defmodule VideoRoom.DisplayEngine do
                 {:forward, {{:endpoint, id}, {:enable_track, new_track_id}}}
               ]
 
-              send(id, {:signal, {:replace_track, old_track_id, new_track_id}})
+              old_participant_id = state.endpoints[old_id].ctx.participant_id
+              new_participant_id = state.endpoints[new_id].ctx.participant_id
+
+              send(id, {:signal, {:replace_track, old_participant_id, new_participant_id}})
```
mickel8, line 103:
```suggestion
              send(id, {:signal, {:replace_participant, old_participant_id, new_participant_id}})
```

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -126,35 +116,25 @@ defmodule VideoRoomWeb.RoomChannel do
...
     push(socket, "participantsList", %{participants: participants})
 
     {:noreply, socket}
   end
 
   @impl true
-  def handle_info({:signal, {:replace_track, old_track_id, new_track_id}}, socket) do
-    push(socket, "replaceTrack", %{
-      data: %{"oldTrackId" => old_track_id, "newTrackId" => new_track_id}
+  def handle_info({:signal, {:replace_track, old_participant_id, new_participant_id}}, socket) do
```
mickel8, line 127:
```suggestion
  def handle_info({:signal, {:replace_participant, old_participant_id, new_participant_id}}, socket) do
```

webrtc/videoroom/lib/videoroom_web/room_channel.ex
```diff
@@ -200,4 +180,28 @@ defmodule VideoRoomWeb.RoomChannel do
...
+        camelized_key
+        |> String.slice(1..-1)
+
+      {
+        key_prefix <> key_sufix,
+        value
+      }
+    end)
+  end
 end
```
mickel8, line 207: Wouldn't be it better to include in this function this piece of code
https://github.com/membraneframework/membrane_demo/blob/ef84c736c0d56120763f068bfa27625c3cf3566b/webrtc/videoroom/lib/videoroom_web/room_channel.ex#L107-L116

and rename it to e.g. prepareJson or serialize etc.

webrtc/videoroom/assets/src/membraneWebRTC.ts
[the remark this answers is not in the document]
mickel8: I am not sure why we have userId nad Prticipant.id. IMO we should have only one id that should be assigned to participant when he joins to the room i.e. here: https://github.com/membraneframework/membrane_demo/blob/master/webrtc/videoroom/lib/videoroom/pipeline.ex#L95

Can't we also use Endpoint.id instead of creating a new UUID.v4()?

webrtc/videoroom/assets/src/room_ui.ts
```diff
@@ -46,24 +46,49 @@ export function getRoomId(): string {
...
-  streamId: string,
+  participantId: string,
   type: "video" | "audio" | "placeholder" | "feed" | "mutedAudioIcon"
 ) {
-  return `${type}-${streamId}`;
+  return `${type}-${participantId}`;
 }
 
-export function addVideoElement(
+export function linkStreamwWithVideoElement(
```
mickel8, line 55: `attachStream`?

mickel8 (changes requested):

mickel8 (approved):

PR #116, opened 2021-06-30
Webrtc to hls debug

Qizot:


webrtc_to_hls/assets/src/ui.ts
```diff
@@ -1,30 +1,15 @@
...
-export function removeVideoElement(_: MediaStreamTrack, stream: MediaStream) {
-  if (stream.getTracks().length > 0) {
-    return;
-  }
-
-  const video = <HTMLVideoElement>document.getElementById(stream.id);
-  if (video) {
-    video.remove();
-  }
+  video.muted = false;
```
FelonEkonom, line 12: When running this app, I have audio feedback - is it caused, by setting this attribute to `false`?

Qizot: Do you mean you can hear your own audio with the streaming view?

PR #126, opened 2021-08-03
use proper reference to RTC engine in the client

mat-hek:


FelonEkonom (approved):

mickel8:
Could we close this and merge #125 as #125  includes also fixes for setting common DTLS key and cert.

PR #127, opened 2021-08-04
bump simple_pipeline dependencies

agga1:


simple_pipeline/mix.exs
```diff
@@ -28,11 +28,11 @@ defmodule Membrane.Demo.SimplePipeline.MixProject do
   # Run "mix help deps" to learn about dependencies.
   defp deps do
     [
-      {:membrane_core, "~> 0.6.0", override: true},
-      {:membrane_file_plugin, "~> 0.5.0"},
-      {:membrane_portaudio_plugin, "~> 0.5.1"},
-      {:membrane_ffmpeg_swresample_plugin, "~> 0.5.0"},
-      {:membrane_mp3_mad_plugin, "~> 0.5.0"},
+      {:membrane_core, "~> 0.7.0", override: true},
```
mickel8, line 31: I think that we can try to get rid of `override: true`
```suggestion
      {:membrane_core, "~> 0.7.0"},
 ```

mickel8 (approved):

simple_pipeline/mix.exs
agga1: ok, I will change that

PR #137, opened 2021-08-31
Send media_event to peer only if channel is not nil

Rados13:
Fix bug which make it possible to send message to `nil`.

It was possible to send message to `nil` in a situation when two peers left room in the same time.

### DIFF
--- a/webrtc/videoroom/lib/videoroom/room.ex
+++ b/webrtc/videoroom/lib/videoroom/room.ex
@@ -53,7 +53,10 @@ defmodule Videoroom.Room do
 
   @impl true
   def handle_info({_sfu_engine, {:sfu_media_event, to, event}}, state) do
-    send(state.peer_channels[to], {:media_event, event})
+    if state.peer_channels[to] != nil do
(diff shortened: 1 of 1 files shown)


webrtc/videoroom/lib/videoroom/room.ex
```diff
@@ -53,7 +53,10 @@ defmodule Videoroom.Room do
 
   @impl true
   def handle_info({_sfu_engine, {:sfu_media_event, to, event}}, state) do
-    send(state.peer_channels[to], {:media_event, event})
+    if state.peer_channels[to] != nil do
```
mickel8, line 56: Can this ever happen? We always call `add_peer_channel` as soon as new WS is created so it shouldn't be possible to trigger WS to emit message to the peer before calling `add_peer_channel`.

Rados13: It is possible in situation when two peers left room in same time and then the message peerLeft is sent to peer which already left.

mickel8 (approved):

PR #139, opened 2021-10-06
Turn api in rtc engine

FelonEkonom:


FelonEkonom:
Done so far: 
 - pass media packets directly between TURN server and RTC Engine via Erlang messages, skipping libNICE

TODO:
 - add support to Firefox
 - remove dependency to libNICE

PR #144, opened 2021-10-13
Adjust to new callbacks removeTrack, updatePeerMetadata, updateTrackMetadata

Rados13:


webrtc/videoroom/assets/src/room.ts
```diff
@@ -42,22 +42,41 @@ export class Room {
...
+                });
+                this.updateParticipantsList();
+              });
+          } catch (error) {
+            console.error(error);
+            setErrorMessage(
+              "Failed to setup video room, make sure to grant camera and microphone permissions"
+            );
+            throw "error";
+          }
```
mickel8, line 70: Would it be possible to put getting localStream in a separate function?
Then we would have
```elixir
    this.localStream = getLocalMedia();
    this.localStream
      .getTracks()
      .forEach((track) => this.webrtc.addTrack(track, this.localStream!));

    addVideoElement(LOCAL_PEER_ID, "Me", true);
    attachStream(this.localStream, LOCAL_PEER_ID);
```

PR #145, opened 2021-10-15
Add fdk-acc plugin

marcinwasowicz:
Could not successfully compile dependencies without installing fdk-acc manually.


marcinwasowicz:
> Looks good! I just noticed that README.md also has an unneeded paragraph "Sample License" (Swan Song is not used in this demo), maybe you could delete it in this pr as well?

Done.

Issue #158, opened 2021-11-10
Make the 'advanced' videoroom demo advanced


mat-hek:
- [x] Move to a separate repo
- [x] Add screen sharing support
- [x] Add ability to disable cam/mic

chavenor:
Let me know when you have this done, I can try to setup it up - I'll clean up the instructions for it as well.  Thanks!

PR #159, opened 2021-11-12
Videoroom advanced controls

Qizot:
Links to #158 
Added controls for muting media, screensharing and leaving the room. Also attached the mute functionality. Screensharing is yet to be done.

image


Rados13 (approved):
It looks and works nice. The only potential thing to change would be to make the disconnect button looks the same in videoroom and advanced_videoroom. But I am not sure about it so it can stay as it is.

PR #161, opened 2021-11-16
Videoroom advanced screensharing

Qizot:
Links to #158 

Added support for screensharing. Adjusted the layout a little bit but it happens to have introduced pillar boxes, it is hard to get it right...

With one screensharing:
image

And we can support more than one 🐺 

image


Rados13 (approved):
Looks and works pretty well.

PR #165, opened 2021-11-26
Merge join() and init() methods in webrtc/videoroom's room.ts into one method

### LINKED ISSUE #164: Merge join and init methods in videoroom demo
I don't really understand an advantage of splitting [Room#init()](https://github.com/membraneframework/membrane_demo/blob/master/webrtc/videoroom/assets/src/room.ts#L82) and [Room#join()](https://github.com/membraneframework/membrane_demo/blob/master/webrtc/videoroom/assets/src/room.ts#L101) into separate functions. This design leads to a weird implementation in [index.ts](https://github.com/membraneframework/membrane_demo/blob/master/webrtc/videoroom/assets/src/index.ts#L4):
``` js
room.init().then(() => room.join());
```
which basically means that `room.join()` and `room.init()` are called synchronously one after another. I couldn't imagine any use case of calling the `room.init()` without calling the `room.join()`, at least in terms of our demo.

If there are no clear disadvantages of merging those functions into one (e.g. simply `join`) I would suggest doing so.

varsill:
closes #164

### DIFF
--- a/webrtc/videoroom/assets/src/index.ts
+++ b/webrtc/videoroom/assets/src/index.ts
@@ -1,4 +1,4 @@
 import { Room } from "./room";
 
-let room = new Room();
+const room = new Room();
 room.join();
--- a/webrtc/videoroom/assets/src/index.ts
+++ b/webrtc/videoroom/assets/src/index.ts
@@ -1,8 +1,4 @@
 import { Room } from "./room";
 
 let room = new Room();
-try {
-  room.join();
-} catch (error) {
-  console.log("error243", error);
-}
+room.join();
--- a/webrtc/videoroom/assets/src/index.ts
+++ b/webrtc/videoroom/assets/src/index.ts
@@ -1,4 +1,8 @@
 import { Room } from "./room";
 
 let room = new Room();
-room.init().then(() => room.join());
+try {
+  room.join();
+} catch (error) {
+  console.log("error243", error);
+}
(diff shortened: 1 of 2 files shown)


webrtc/videoroom/assets/src/room.ts
```diff
@@ -79,7 +79,21 @@ export class Room {
     );
   }
 
-  public init = async () => {
+  public join = () => {
+    this.init()
```
Qizot, line 83: Can't you make the join function async and then use `await` and `try/catch` syntax instead of `then` and `catch`? Promises in JS are eager instead of being lazy so it will eventually execute, you just have to use await if you want it to finish executing before certain point in code. So calling `room.join` will behave the same being and not being asynchronous.

Qizot (changes requested):

webrtc/videoroom/assets/src/room.ts
varsill: Sure, makes sense, I will change this

Qizot (approved):

webrtc/videoroom/assets/src/index.ts
```diff
@@ -1,4 +1,4 @@
 import { Room } from "./room";
 
 let room = new Room();
```
Qizot, line 3: Can you change `let` to `const`?

varsill: sure ;) done

PR #170, opened 2021-12-07
Remove hls dependencies

Rados13:


webrtc/videoroom/mix.exs
```diff
@@ -23,7 +23,7 @@ defmodule VideoRoom.MixProject do
     [
       {:membrane_core, github: "membraneframework/membrane_core", override: true},
       {:esbuild, "~> 0.1", runtime: Mix.env() == :dev},
-      {:membrane_rtc_engine, github: "membraneframework/membrane_rtc_engine"},
+      {:membrane_rtc_engine, github: "membraneframework/membrane_rtc_engine", branch: "optional"},
```
Qizot, line 26: I guess the `optional` branch is already merged and this will fail?

Qizot (changes requested):

webrtc/videoroom/mix.exs
Rados13: Yes you are right.

mickel8:
I have already pushed this to master :roll_eyes:

PR #171, opened 2021-12-08
Added naive simulcast

Rados13:
- modify webrtc_to_hls demo to use new modular_rtc_engine,
- webrtc_to_hls demo use simulcast by default


webrtc/videoroom/priv/static/js/room.js
```diff
@@ -0,0 +1,231 @@
+/******/ (function(modules) { // webpackBootstrap
```
mickel8: This probably should be in `.gitignore`

webrtc_to_hls/assets/src/app.ts
```diff
@@ -27,7 +27,7 @@ export const AUDIO_CONSTRAINTS: MediaStreamConstraints = {
 
 export const VIDEO_CONSTRAINTS: MediaStreamConstraints = {
   audio: false,
-  video: { width: 640, height: 360, frameRate: 24 },
+  video: { width: 1280, height: 720, frameRate: 24 },
```
mickel8, line 30: Why do we want to send HD video? This will take more bandwidth and CPU resources

webrtc/videoroom/assets/src/consts.ts
```diff
@@ -1,6 +1,6 @@
 export const MEDIA_CONSTRAINTS: MediaStreamConstraints = {
   audio: true,
-  video: { width: 640, height: 360, frameRate: 24 },
+  video: { width: { exact: 1280 }, height: { exact: 720 }, frameRate: 24 },
```
mickel8, line 3: Why HD?

webrtc/videoroom/assets/src/room.ts
```diff
@@ -66,6 +66,7 @@ export class Room {
           addVideoElement(peer.id, peer.metadata.displayName, false);
         },
         onPeerLeft: (peer) => {
+          console.log("Peer left");
```
mickel8, line 69: We don't have any other logs, about peer joining etc. so maybe let's be consistent and remove it?

webrtc/videoroom/priv/static/js/room.js
mickel8: This should be in .gitignore. Please check how does `.gitignore` look like in other examples

mickel8 (changes requested):

webrtc_to_hls/assets/src/app.ts
Rados13: When I was developing this, somewhere I found information that HD is required for simulcast and only with HD constraints it was working. But as I am testing this now it can be decreased to previous constraints.

Rados13: Ok I remember on beginning I was trying to get 3 simulcast layers and based on this [discussion](https://groups.google.com/g/discuss-webrtc/c/czRbYBm1u2g/m/xgawiGCcDgAJ), I thought that HD is required. The correct limits are [here](https://chromium.googlesource.com/external/webrtc/+/master/media/engine/simulcast.cc#89).

mickel8 (approved):
There is still one comment to resolve. Besides this I think we can mereg but please merge webrtc_plugin at first

webrtc/videoroom/priv/static/js/room.js
Rados13: Hmm they are looking the same

PR #172, opened 2021-12-15
OpenTelemetry

Rados13:
For now, three Opentelemetry exporters are supported:
- stdout exporter - print all spans in terminal
- honeycomb - to use it you have to specify the Honeycomb API key under HONEYCOMB environment variable
- Zipkin  - you can run Zipkin service on a local machine with the command `docker run -d -p 9411:9411 openzipkin/zipkin`


Qizot (approved):
It's pretty bad that we have to duplicate the same code for the simple and advanced version but I guess nothing can be done about that.

mat-hek:
Not sure if we want to add telemetry to the basic videoroom. The intention was to keep it as simple as possible @Rados13 @mickel8

PR #173, opened 2021-12-23
Fix simultaneous joining to the room

mickel8:


varsill (approved):
Seems legit.
It's nice that you have found out that gen_server returns such an error message code :)

PR #191, opened 2022-02-21
Adjust to simulcast encoding switch implementation in RTC Engine

mickel8:


mickel8:
Moved to membraneframework/membrane_videoroom#9

PR #198, opened 2022-04-04
Feature/camera to hls

dimamik:


camera_to_hls/README.md
```diff
@@ -0,0 +1,52 @@
+# Membrane Demo - Camera Video to HLS
+
+This project demonstrates capturing camera video and converting it to HLS stream, which is then can be used wherever you want :)
+
+## Prerequisites
+
+You must have following packages installed on your system:
+
+- ffmpeg 4.\*
```
daniel-jodlos, line 9: I think openssl is also required. Native dependencies probably won't compile without it

camera_to_hls/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Run the demo
+
+Run the project with:
+
+```bash
+mix run --no-halt
+```
+
+HLS header and segment files will be created in `output` directory along with playlist files.
```
daniel-jodlos, line 25: That's incorrect. HLS only defines a playlist. Headers and segments are CMAF. If you want to know the boundary of HLS and what is something else, checkout [RFC 8216](https://datatracker.ietf.org/doc/html/rfc8216)

camera_to_hls/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+To play the HLS stream you need to serve the content of `output` dir, e.g. by running:
```
daniel-jodlos, line 27: I think I would rephrase it to highlight that HLS can be served via a regular HTTP server.

camera_to_hls/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Then, you can open the url http://localhost:8000/output/index.m3u8 in some player, e.g. `ffplay` or `vlc`
+
+```bash
+ffplay http://localhost:8000/output/index.m3u8
+```
+
+Moreover you can open the url http://localhost:8000/stream in your browser and enjoy the video from your camera :)
+
+_You would be asked to grant access to your camera, and after that you can enjoy the video from your camera_
```
daniel-jodlos, line 41: For me this sentence doesn't match well with the previous one styling wise

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -0,0 +1,51 @@
...
+  @impl true
+  def handle_init(_) do
+    children = [
+      # Captures video from the camera (raw video, depending on camera/os)
+      source: Membrane.CameraCapture,
+      # Converts pixel format to I420 (this is still a raw video, but different format)
+      converter: %Membrane.FFmpeg.SWScale.PixelFormatConverter{format: :I420},
+      # Takes raw I420 frames and converts to H264
+      # baseline profile tells encoder not to generate
+      # B-frames because browsers can't handle them
```
daniel-jodlos, line 13: First problem: grammatical correctness. Second issue is that these sentences are not precise enough.
```suggestion
      # Takes raw video in I420 pixel format and encodes it into H264.
      # Baseline profile tells encoder not to generate
      # B-frames because browsers can't render them properly
```

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -0,0 +1,51 @@
...
+      encoder: %Membrane.H264.FFmpeg.Encoder{profile: :baseline},
+      # Creates caps, generates metadata and adds timestamps
+      # There we need to specify framerate, based on it we
+      # generate timestamps
+      # Further we won't be able to control framerate
```
daniel-jodlos, line 18: That information about controlling framerate is:
1. probably unnecessary
1. confusing. We don't really control the framerate here, we just generate timestamps based on it

Moreover, I would rephrase it
```suggestion
      # Creates caps, generates metadata and timestamps the stream based on the framerate
```

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -0,0 +1,51 @@
...
+      video_nal_parser: %Membrane.H264.FFmpeg.Parser{
+        framerate: {30, 1},
+        alignment: :au,
+        attach_nalus?: true
+      },
+      # Generates caps for mp4 by changing H264
+      # It is the same H264, only difference is the way of division
```
daniel-jodlos, line 25: Changing H264? You're not being precise. H264 is converted for Annex B form to length prefixed form. This is called delimitation, not `the way of division`. Also, this is quite important, this elements payloads the H264 so that it can be injected into MP4

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -0,0 +1,51 @@
...
+      video_payloader: Membrane.MP4.Payloader.H264,
+      # There we perform CMAF muxing
+      # Simply put it generates containers of specified length
+      # Which than can be used in HLS or other transporting protocols
```
daniel-jodlos, line 29: I wouldn't mention other protocols.

While we're at it, not other transporting protocols, I would say it smells like a Transport Layer Protocol from ISO/OSI. HTTP Streaming procotols, namely, MPEG-DASH, be we don't implement it, so I would skip that

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -0,0 +1,51 @@
...
+      video_nal_parser: %Membrane.H264.FFmpeg.Parser{
+        framerate: {30, 1},
+        alignment: :au,
+        attach_nalus?: true
+      },
+      # Generates caps for mp4 by changing H264
+      # It is the same H264, only difference is the way of division
+      video_payloader: Membrane.MP4.Payloader.H264,
+      # There we perform CMAF muxing
+      # Simply put it generates containers of specified length
```
daniel-jodlos, line 28: Segments, not containers. Their are container if you prepend the header to it. Otherwise, they are CMAF segments

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -0,0 +1,51 @@
...
+      # Which than can be used in HLS or other transporting protocols
+      # to transport data in chunks
+      video_cmaf_muxer: Membrane.MP4.Muxer.CMAF,
+      # HLS element is responsible for generating HTTP streaming manifests.
```
daniel-jodlos, line 32: HLS playlists

daniel-jodlos (changes requested):

daniel-jodlos (commented):
I've found some typos but unfortunately there is still some more serious issues that preclude approval.

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -8,28 +8,28 @@ defmodule Membrane.Demo.CameraToHls.Pipeline do
...
+      # Also, we generate the timestamps based on the framerate specified
       video_nal_parser: %Membrane.H264.FFmpeg.Parser{
         framerate: {30, 1},
         alignment: :au,
         attach_nalus?: true
       },
       # Generates caps for mp4 by changing H264
-      # It is the same H264, only difference is the way of division
+      # H264 is converted for Annex B form to length prefixed form,
+      # this process is called delimitation
```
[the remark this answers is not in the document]
dimamik: Yup, didn't get this in previous review. Now everything is clear, thanks!

camera_to_hls/lib/membrane_demo_camera_to_hls/pipeline.ex
```diff
@@ -8,28 +8,28 @@ defmodule Membrane.Demo.CameraToHls.Pipeline do
...
+      # Also, this element payloads the H264 so that it can be injected into MP4
```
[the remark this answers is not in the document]
dimamik: I think that for the sake of this demo this parts can be omitted to simplify the flow, but thanks a lot for mentioning, I'll definitely get deeper into that.

daniel-jodlos (approved):

PR #216, opened 2022-09-19
Updated simple element demo.

shuntrho:


simple_element/README.md
```diff
@@ -24,16 +23,25 @@ The pipeline is taken from the `simple_pipeline` demo.
 1. Make sure you have Elixir installed on your machine. See: https://elixir-lang.org/install.html
 1. Fetch the required dependencies by running `mix deps.get`
```
varsill, line 24: I believe we should make the prerequisites section more explicit, i.e. I would explicitly say, that the user needs to run `mix.deps.get` inside the project file, or I would move that point at the beggining of the `Run the demo" section, where we are assuming, that the user runs the code inside the project directory.

And another funny thing (Probably not a bug though) In the "Files changed" view I see that the enumeration is wrong (I see numer `1.` twice in a row) :D

varsill (approved):
I have just left one thing to consider

simple_element/README.md
shuntrho: I will leave it as it is, I assume that it rather obvious that you need to be in the demo's directory.

PR #217, opened 2022-09-21
Update and fix rtp demo

shuntrho:
Updated deps and moved audio and video samples to the repo.


rtp/lib/membrane_demo_rtp/send_pipeline.ex
```diff
@@ -33,34 +36,35 @@ defmodule Membrane.Demo.RTP.SendPipeline do
           ]
         },
         video_realtimer: Membrane.Realtimer,
-        video_sink: %Membrane.Element.UDP.Sink{
+        video_sink: %Membrane.UDP.Sink{
```
varsill, line 39: NIT: how about adding some aliases, so that not to have to write `Membrane.*` each time?

rtp/README.md
```diff
@@ -32,11 +32,8 @@ You should be able to see an SDL player showing an example video.
 The stream can also be sent with GStreamer:
```
varsill, line 32: Hmm I believe we should prepare more complex readme - 
i.e. we should explicitly say which part of the system does the gstream substitue, as well as I would provide some links to the dependencies that need to be installed (even to fixed versions of these dependencies). We can also think about adding a brief description of what is going on within `receive.exs` and `send.exs`.

varsill (approved):
lgtm! 🥇

PR #219, opened 2022-11-10
Add instruction to set `Keyframe Interval` in rtmp2hls readme

daniel-jodlos:
A README for `rtmp_to_hls` doesn't intruct a user to set `Keyframe Interval` in OBS. Given that output is created by HLS, periodical keyframes are required to correctly split the stream into CMAF segments.

Lack of that option will result in incorrect lengths of CMAF segments. As a special case of this, the start of the HLS stream will be delayed, as the very first segment will not appear until the encoder on the sender side decides to generate a keyframe.


varsill (approved):
LGTM 🥇 
Let's merge it the way it is and we will update the video in the README.md later on (Or we will remove it because I'm not sure if anybody considers it to be helpful anyhow :D)

PR #222, opened 2022-11-16
Updated dependencies.

shuntrho:


varsill (approved):

varsill:
I believe you should do fix `mix.lock` by using `mix deps.update --all` in the project dir - after that we can merge it.

PR #223, opened 2022-11-25
Pure elixir rtmp source

varsill:
This PR makes the RTMP to HLS demo use pure elixir RTMP Source. A HLS's `:target_window_duration` has been set to `:infinity` so that to support live streams.
Furthermore, small improvements the README.md for that demo have been made.
NOTE: we need to get rid of github dependencies in the mix.exs, but it will be possible once `membrane_rtmp_plugin` is released with Core v0.11. Then we will also need to update the Core version in the RTMP to HLS demo.


DominikWolek (changes requested):
Please delete the `DS_Store`. Other than that, looks good to me :D

DominikWolek (approved):

PR #224, opened 2022-11-29
Add instructions to set `EXTERNAL_IP` when running with mix.

jerzywilczek:
Without setting this, the app will not work in FIrefox.


varsill (commented):
In the membrane_videoroom repository there is a more comprehensive description on how to receive that address - we can consider linking to that description:
https://github.com/membraneframework/membrane_videoroom#launching-of-the-application-1

varsill (approved):

PR #249, opened 2023-10-20
Add speech to text livebook

mat-hek:


speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+  ],
+  config: [
+    nx: [default_backend: EXLA.Backend]
+  ]
+)
+```
+
+## Prerequisites
+
+On Linux, you need to install PortAudio, for example with `sudo apt install portaudio19-dev`.
```
varsill, line 21: I would point to the installation section in the `membrane_portaudio_plugin`'s README (the manual there covers multiple OS and "is aware" of precompiled deps)

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+
+## Introduction
+
+This livebook example shows how to perform a real-time speech-to-text conversion with the use of the [Membrane Framework](https://github.com/membraneframework) and the [Bumblebee](https://github.com/elixir-nx/bumblebee).
+
+You will see how to fetch the audio from your microphone, perform preprocessing and create your own Membrane element that runs an AI speech-to-text conversion model.
+
+## Element performing a speech to text
```
varsill, line 29:
```suggestion
## Element performing speech to text conversion
```

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+
+We need to write a custom sink that will perform speech-to-text and display the resulting transcription. Let's call this element `SpeechToText`.
```
varsill, line 31: Let's link for instance there: https://membrane.stream/learn/glossary#sink
```suggestion
We need to write a custom [sink](https://membrane.stream/learn/glossary#sink) that will perform speech-to-text and display the resulting transcription. Let's call this element `SpeechToText`.
```

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+spec =
+  child(%Membrane.PortAudio.Source{channels: 1, sample_format: :f32le, sample_rate: 16_000})
+  |> child(SpeechToText)
+  |> child(%Membrane.Debug.Sink{handle_buffer: &IO.puts(&1.payload)})
+
+pipeline = RCPipeline.start_link!()
+RCPipeline.exec_actions(pipeline, spec: spec)
+```
+
+The prints with the transcription should start appearing below the last cell you have evaluated.
```
varsill, line 210: #How about warning the user that he might be asked for permissions to use microphone?

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+
+The Whipser model requires the input audio samples to be in `f32le` format, which means, that they are represented as floating numbers written on 32 bits, with little endian bytes order. The required input sample rate is 16 kHz, as well as only single-channel samples are allowed.
+
+### Initialization
+
+In the initialization process, we load the `Whisper` model, along with the featurizer, tokenizer, and generation configuration. Then we create a speech-to-text serving and indicate that we want to use `EXLA` backend for the tensor operations.
+
+### Buffers handling
+
+Once a buffer arrives, we check if it contains enough amount of audio and accumulate it in the state otherwise. Then we perform basic VAD (Voice Activity Detection) to remove silent chunks, what improves the model behaviour. Then we convert the audio to an Nx tensor, feed the model and parse output, which is then sent to via the output pad.
```
varsill, line 44: Let's link to `Nx`

varsill, line 44:
```suggestion
Once a buffer arrives, we check if it contains enough amount of audio and accumulate it in the state otherwise. Then we perform basic VAD (Voice Activity Detection) to remove silent chunks, which improves the model behaviour. Then we convert the audio to an Nx tensor, feed the model, and parse the output, which is then sent via the output pad.
```

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+
+### Problems and potential improvements
```
varsill, line 46: How about putting it at the bottom of the notebook?

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+We welcome ideas and contributions addressing these problems or bringing other improvements.
+
+```elixir
+defmodule SpeechToText do
+  use Membrane.Filter
+
+  alias Membrane.RawAudio
+  require Membrane.Logger
+
+  # 8000 4-byte samples, at 16kHz it's half a second
```
varsill, line 61: How about setting the time (in this case 500ms) as an attribute and calculating the `vad_chunk_size` based on that? (I believe we can use (that)[https://github.com/membraneframework/membrane_raw_audio_format/blob/49d5dd11717d872edb377aa5afa26f2392225b7c/lib/membrane_raw_audio.ex#L353) function for this purpose)

speech_to_text/speech_to_text.livemd
varsill: I have just realized that the `SpeechToText` is not a sink, but a filter :D (I believe there are some other places where we refer to the `SpeechToText` element as "sink")

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+  defp generate_chunks(samples) do
+    [samples]
+  end
+
+  defp calc_volume(chunk) do
+    samples = for <<sample::float-32-little <- chunk>>, do: sample
+    samples_cnt = Enum.count(samples)
+    samples_avg = Enum.sum(samples) / samples_cnt
+    sum_mean_square = samples |> Enum.map(&((&1 - samples_avg) ** 2)) |> Enum.sum()
+    :math.sqrt(sum_mean_square / samples_cnt)
```
varsill, line 184: [NIT] We could probably delegate that part to some kind of `calculate_standard_deviation` function and describe why standard deviation represents volume

speech_to_text/speech_to_text.livemd
```diff
@@ -0,0 +1,217 @@
...
+  end
+
+  defp filter_silence(samples, state) do
+    samples
+    |> generate_chunks()
+    |> Enum.drop_while(&(calc_volume(&1) < state.vad_threshold))
+    |> Enum.join()
+  end
+
+  defp generate_chunks(samples) when byte_size(samples) >= 2 * @vad_chunk_size do
```
varsill, line 170: Are you sure it shouldn't be written that way?
```suggestion
  defp generate_chunks(samples) when byte_size(samples) >= @vad_chunk_size do
```

varsill (changes requested):

speech_to_text/speech_to_text.livemd
mat-hek: This is on purpose. I assumed it's better to have slightly longer chunks sometimes than leftovers shorter than `vad_chunk_size`

speech_to_text/speech_to_text.livemd
mat-hek: Added a comment

varsill (approved):
LGTM 🥇

PR #263, opened 2023-12-14
[camera_to_hls_nerves] Stop running `:inets` http server in a GenServer

Noarkhh:


camera_to_hls_nerves/mix.exs
```diff
@@ -37,8 +37,7 @@ defmodule CameraToHlsNerves.MixProject do
       {:ring_logger, "~> 0.10.0"},
       {:toolshed, "~> 0.3.0"},
 
-      {:membrane_rpicam_plugin, "~> 0.1.2"},
-      {:membrane_h264_plugin, "~> 0.9.0"},
```
mat-hek, line 41: It seems we use the h264 plugin directly, so it should be in the deps too

camera_to_hls_nerves/lib/camera_to_hls_nerves/hls_server.ex
```diff
@@ -7,16 +7,24 @@ defmodule CameraToHlsNerves.HlsServer do
...
-    :inets.start(:httpd, httpd_options())
+    {:ok, server} = :inets.start(:httpd, httpd_options())
+
+    {:noreply, %{state | server: server}}
+  end
+
+  @impl true
+  def terminate(_reason, state) do
+    :inets.stop(:httpd, state.server)
+    state
```
mat-hek, line 27: The problem with the `terminate` callback is that it's called only in particular cases. In short, to work reliably the GenServer would have to trap exits (which would complicate implementation) and not have bugs :D because if the GenServer crashes, terminate is never called.

After reading this again, I think we don't need this GenServer. We can either
- Start the server from within the `Application.start` [here](https://github.com/membraneframework/membrane_demo/blob/fix-camera-to-hls-nerves/camera_to_hls_nerves/lib/camera_to_hls_nerves/application.ex). It won't be under the same supervisor as the pipeline, but I'm not sure if it matters
- Use the [standalone](https://www.erlang.org/doc/man/inets.html#start-3) mode of the server, which doesn't require `:inets.start()` and allows for starting the server under a specific supervisor, like
```elixir
Supervisor.start_link(
  [%{id: :hls_server, start: {:inets, :start, [:httpd, opts, :stand_alone]}}],
  strategy: :one_for_one
)
```

mat-hek (changes requested):

camera_to_hls_nerves/lib/camera_to_hls_nerves/hls_server.ex
Noarkhh: If we use the second method we lock ourselves out of passing the `:playlist_ready` message to start the server, but I think we don't really need it, since the demo works without it

mat-hek: I forgot about it. But instead of spawning the server on startup, we can do `Supervisor.start_child` at any point

mat-hek (approved):
As mentioned in the comment, we can use `Supervisor.start_child` to delay the spawning of the HTTP server. But if it works, you can leave it as is.

camera_to_hls_nerves/lib/camera_to_hls_nerves/hls_server.ex
Noarkhh: After looking some more the `:playlist_ready` ensured that the server would always have the stream already loaded, and without it a refresh is sometimes needed

Noarkhh: Also I'm not sure where `Supervisor.start_child` could be invoked, since it would need the top supervisor's pid, I'm not sure if some solution using registration is a good practice

mat-hek: I think name registration is quite common in such cases. It's as simple as passing the `name` option to the supervisor and using it in the pipeline

mat-hek (approved):

PR #269, opened 2024-02-10
Add work around to CameraCapture plugin error

guissalustiano:
Hi,
I was following the tutorials and found a problem that I managed to fix, maybe this information can be useful for other people.
I'm new to video encoding and the membrane framework so maybe there is a better way or explanation to fix that.


mat-hek (approved):

mat-hek:
Thanks!

PR #270, opened 2024-02-12
Livebooks cleanup

bartkrak:
- bump versions of all packages
- change mp3 sample
- remove native deps instalation


mat-hek:
Let's remove https://github.com/membraneframework/membrane_demo/blob/livebooks-cleanup/livebooks/assets/path_set.png since it's not needed

mat-hek (approved):

PR #271, opened 2024-02-12
refactor soundwave livebook

mat-hek:


livebooks/soundwave/soundwave.livemd
```diff
@@ -68,134 +56,96 @@ defmodule Visualizer do
...
+      |> Vl.encode_field(:x, "x", title: "Time [s]", type: :quantitative)
+      |> Vl.encode_field(:y, "y",
+        title: "Amplitude",
+        type: :quantitative,
+        scale: %{domain: [-1.1, 1.1]}
+      )
+      |> Kino.VegaLite.new()
+
+    Kino.render(chart)
+    chart
```
varsill, line 148: [NIT] It should be fine as `Kino.render/1` returns the `chart` 
```suggestion
    Kino.render(chart)
```

livebooks/soundwave/soundwave.livemd
```diff
@@ -68,134 +56,96 @@ defmodule Visualizer do
...
+    state = if state.pts == nil, do: %{state | pts: buffer.pts}, else: state
+    stream_format = ctx.pads.input.stream_format
+    sample_size = RawAudio.sample_size(stream_format)
+    sample_max = RawAudio.sample_max(stream_format)
 
     samples =
-      for <<sample::binary-size(state.bytes_per_sample) <- buffer.payload>> do
-        RawAudio.sample_to_value(sample, ctx.pads.input.stream_format)
+      for <<sample::binary-size(sample_size) <- buffer.payload>> do
+        RawAudio.sample_to_value(sample, stream_format) / sample_max
```
varsill, line 95: Shouldn't it be more like: 
```suggestion
        RawAudio.sample_to_value(sample, stream_format) / (sample_max-sample_min)
```?
(I am a little bit worried it won't get rescaled properly in case we have a signed number as a sample value)

mat-hek: I think PortAudio returns signed numbers by default. `sample_max` is around `2^15` and `sample_min` is around `-(2^15)`, so by dividing by `sample_max` we're getting `[-1, 1]`. After the proposed change, I think we'd get `[-0.5, 0.5]` 🤔

varsill (approved):

PR #277, opened 2024-04-18
Bump dependencies in rtmp_to_hls demo

FelonEkonom:
Deps bumped to the versions present in the newest `Membrane SDK`.
For some reason browser page at localhost:4000 seems to not render something, `npm install` in `assets` doesn't help


mat-hek (approved):
Fix or remove the logo url and you can merge

PR #280, opened 2024-06-14
Use rtsp plugin

Noarkhh:


rtsp_to_hls/lib/pipeline.ex
mat-hek, line 1: Since you're touching this, let's remove the `if` in [line 64](https://github.com/membraneframework/membrane_demo/pull/280/files#diff-a3620d32d19b5d952cff711fefd834f7005ad6de9ad1225299339f72eef2dac0R64) and pattern match instead ;)

PR #281, opened 2024-07-19
Update README.md

harrisi:
`start_link` returns a triple. A two-element tuple still works (i.e., plays the audio), but it throws a `MatchError`.


FelonEkonom (approved):
LGTM, thank you very much for your contribution 👍

PR #283, opened 2024-08-26
Update the rtsp_to_hls demo

Noarkhh:


rtsp_to_hls/lib/pipeline.ex
```diff
@@ -63,34 +84,22 @@ defmodule Membrane.Demo.RtspToHls.Pipeline do
           generate_best_effort_timestamps: %{framerate: {30, 1}}
```
mat-hek, line 84: is that needed?

rtsp_to_hls/lib/server/handler.ex
mat-hek, line 1: I'm wondering if we should put the server code in the `lib` directory since this demo is called rtsp_to_hls 🤔 Maybe we should keep all the server stuff in `server.exs`?

rtsp_to_hls/lib/pipeline.ex
```diff
@@ -13,72 +13,65 @@ defmodule Membrane.Demo.RtspToHls.Pipeline do
...
-        {:new_track, ssrc, %{type: :video} = track},
-        :source,
-        _ctx,
-        %{rtp_started: false} = state
-      ) do
-    Logger.debug(":new_rtp_stream")
+  def handle_child_notification({:set_up_tracks, tracks}, :source, _ctx, state) do
+    tracks_left_to_link =
+      [:audio, :video]
+      |> Enum.filter(fn media_type -> Enum.any?(tracks, &(&1.type == media_type)) end)
```
mat-hek, line 35: So we just support one audio and one video track? If so, I'd just add `audio_track` and `video_track` fields to the state, it should be much simpler

rtsp_to_hls/lib/pipeline.ex
```diff
@@ -13,72 +13,65 @@ defmodule Membrane.Demo.RtspToHls.Pipeline do
...
+            spec: [
+              child(:hls, %Membrane.HTTPAdaptiveStream.SinkBin{
+                target_window_duration: Membrane.Time.seconds(120),
+                manifest_module: Membrane.HTTPAdaptiveStream.HLS,
+                storage: %Membrane.HTTPAdaptiveStream.Storages.FileStorage{
+                  directory: state.output_path
+                }
+              })
+              | track_specs
+            ]
```
mat-hek, line 58:
```suggestion
            hls =
              child(:hls, %Membrane.HTTPAdaptiveStream.SinkBin{
                target_window_duration: Membrane.Time.seconds(120),
                manifest_module: Membrane.HTTPAdaptiveStream.HLS,
                storage: %Membrane.HTTPAdaptiveStream.Storages.FileStorage{
                  directory: state.output_path
                }
              })

            [spec: [hls | track_specs]]
```

rtsp_to_hls/lib/pipeline.ex
```diff
@@ -13,72 +13,65 @@ defmodule Membrane.Demo.RtspToHls.Pipeline do
...
-
-    {[spec: structure], %{state | rtp_started: true}}
+    {[], %{state | tracks_left_to_link: tracks_left_to_link}}
   end
 
   @impl true
-  def handle_child_notification({:new_track, ssrc, _track}, :source, _ctx, state) do
-    Logger.warning("new_rtp_stream Unsupported stream connected")
+  def handle_child_notification({:new_track, ssrc, track}, :source, _ctx, state) do
+    if Enum.member?(state.tracks_left_to_link, track.type) do
```
mat-hek, line 42:
```suggestion
    if track.type in state.tracks_left_to_link do
```

mat-hek (approved):

rtsp_to_hls/lib/pipeline.ex
Noarkhh: I don't really think so, since we support also only video (and only audio) streams, I tried doing this and a lot of branching arises

PR #287, opened 2024-10-28
Add RTMP Source to adaptive bitrate HLS example

lastcanal:
Hello! This is a demo of a Pipeline that converts an RTMP stream into adaptive bit-rate HLS files and serves it with Vidstack/HLS.js.

This demo also serves as a showcase for an API proposal for https://github.com/membraneframework/membrane_rtmp_plugin/pull/103 , which sends RTMP Messages to the RTMP client handler. 

Here is a list of the Membrane plugins that are demoed:

```elixir
{:membrane_framerate_converter_plugin, "~> 0.8.2"},
{:membrane_ffmpeg_swscale_plugin, "~> 0.15.1"},
{:membrane_h264_ffmpeg_plugin, "~> 0.31.6"},
{:membrane_http_adaptive_stream_plugin, "~> 0.18.5"},
{:membrane_rtmp_plugin, "~> 0.26.0"},
{:membrane_tee_plugin, "~> 0.12.0"}

```


rtmp_to_adaptive_hls/lib/rtmp_to_adaptive_hls/pipeline.ex
```diff
@@ -0,0 +1,100 @@
...
+
+    {[spec: structure], %{client_ref: client_ref}}
+  end
+
+  @impl true
+  def handle_info(
+        {:metadata_message, %Membrane.RTMP.Messages.SetDataFrame{} = message},
+        _ctx,
+        state
+      ) do
```
varsill, line 43: I think it would be good to make sure that this callback doesn't get called more than once, which could happen if `SetDataFrame` is sent multiple times (I don't think RTMP disallows that)

rtmp_to_adaptive_hls/lib/rtmp_to_adaptive_hls/pipeline.ex
```diff
@@ -0,0 +1,100 @@
...
+
+  @impl true
+  def handle_info(_message, _ctx, state) do
+    {[], state}
+  end
+
+  @impl true
+  def handle_call(:get_video_id, _ctx, state) do
+    {[{:reply, state.video.id}], state}
+  end
```
varsill, line 93: Do we need that `:get_video_id`? I also don't think it would work, as there is no `:video` field in the pipeline's state

varsill (changes requested):
Thank you very much for the contribution! 
I've left some comments, please let me know if something is unclear ;)

rtmp_to_adaptive_hls/doc_assets/RTMP_to_HLS_pipeline.png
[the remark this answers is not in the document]
lastcanal: The README has been updated, hopefully it explains the 'adaptive' part.

rtmp_to_adaptive_hls/priv/static/robots.txt.gz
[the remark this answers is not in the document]
lastcanal: I've removed these and a number of other unneeded files.

rtmp_to_adaptive_hls/lib/rtmp_to_adaptive_hls/pipeline.ex
lastcanal: This is handled now by storing the metadata message in the client handler's state and overloading handle_info/3 to only get called if the metadata message has yet to arrive.

rtmp_to_adaptive_hls/lib/rtmp_to_adaptive_hls/pipeline.ex
lastcanal: Nope! This is an artifact from the `rtmp_to_hls` demo.

varsill (commented):
Hello! Great, thanks for implementing suggestions.
I believe that you need to revoke `assets` directory because currently looks like that:
image

Apart from that, I've slightly modified the diagram:
image - could you exchange the current one in the `doc_assets` directory with that one?

rtmp_to_adaptive_hls/doc_assets/RTMP_to_HLS_pipeline.png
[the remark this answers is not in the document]
varsill: Great, it indeed does!

lastcanal:
Guess I went a little overboard removing assets! I've dropped that commit and forced-pushed with a new commit. It looks good again! Thanks for the updated diagram, looks great!

varsill (approved):
Thank you very much for the contribution! 🥇

PR #291, opened 2025-04-02
Webrtc LiveView

FelonEkonom:


webrtc_live_view/lib/webrtc_live_view_web/live/home.ex
```diff
@@ -0,0 +1,47 @@
+defmodule WebrtcLiveViewWeb.Live.EchoLive do
+  use WebrtcLiveViewWeb, :live_view
+
+  alias Membrane.WebRTC.Live.{Capture, Player}
+
+  def mount(_params, _session, socket) do
+    socket =
+      if connected?(socket) do
```
varsill, line 8: When can it happen, that the socket is not connected in `mount`?

webrtc_live_view/config/config.exs
```diff
@@ -34,7 +34,8 @@ config :esbuild,
 # Configures Elixir's Logger
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
-  metadata: [:request_id]
+  metadata: [:request_id],
+  level: :info
```
varsill, line 38: Why `:info`?

webrtc_live_view/assets/js/app.js
```diff
@@ -16,29 +16,38 @@
...
+import { LiveSocket } from "phoenix_live_view";
+import topbar from "../vendor/topbar";
+import { createCaptureHook, createPlayerHook } from "membrane_webrtc_plugin";
 
-let csrfToken = document.querySelector("meta[name='csrf-token']").getAttribute("content")
+const iceServers = [{ urls: "stun:stun.l.google.com:19302" }];
+
+let Hooks = {};
+Hooks.Capture = createCaptureHook(iceServers);
+Hooks.Player = createPlayerHook(iceServers);
```
varsill, line 30: Do we need to call it `Hooks` (and write it in PascalCase)?

webrtc_live_view/lib/webrtc_live_view/contours_drawer.ex
```diff
@@ -0,0 +1,45 @@
+defmodule WebRTCLiveView.CountoursDrawer do
+  use Membrane.Filter
+
+  alias Evision.{ColorConversionCodes, Constant}
+
+  def_input_pad :input, accepted_format: Membrane.RawVideo
```
varsill, line 6: You could provide pixel format here as well

webrtc_live_view/lib/webrtc_live_view/contours_drawer.ex
```diff
@@ -0,0 +1,45 @@
...
+    {contours, _} =
+      Evision.findContours(
+        bw,
+        Constant.cv_RETR_LIST(),
+        Constant.cv_CHAIN_APPROX_NONE()
+      )
+
+    contours =
+      Enum.filter(contours, fn c ->
+        trunc(Evision.contourArea(c)) in 100..200_000
```
varsill, line 31: What is that purpose of that filtering?

webrtc_live_view/lib/webrtc_live_view_web/live/home.ex
FelonEkonom: `mount/3` is always called 2 times, the first time when the socket is not connected, the second time when it is connected

webrtc_live_view/lib/webrtc_live_view/contours_drawer.ex
FelonEkonom: IDK, I took this code from @mat-hek

webrtc_live_view/config/config.exs
FelonEkonom: :debug makes the console look horrible

webrtc_live_view/lib/webrtc_live_view/contours_drawer.ex
varsill: These are some magic numbers but I don't think there is any value in assigning them names since we don't know if they have any semantic sense, so perhaps let's leave it as it is :D

